Issue #41, opened 2026-02-12
Gardener SLO monitoring


etiennnr:
### Enhancement Description

- One-line enhancement description: An SLO (service level objective) extension that provides a standardized framework for measuring, monitoring and alerting on shoot cluster reliability from the customer perspective.
- Steering: https://gardener.cloud/community/steering/0041-slo-monitoring/
- Gardener Enhancement Proposal: https://github.com/gardener/enhancements/pull/42
- Tracking Issue:  n/a
- GEP (`g/enhancements`) update PR(s):

timebertt:
ref https://gardener.cloud/community/steering/0041-slo-monitoring/ and https://github.com/gardener/enhancements/pull/42#issuecomment-4659547682
/close

PR #42, opened 2026-02-12
GEP-0041: SLO monitoring proposal

etiennnr:
- One-line PR description: Adding proposal for SLO monitoring previously made in https://github.com/gardener/documentation/pull/818 (adding it here since process changed)

- Issue link: #41 

- Other comments: n/a


geps/0041-slo-monitoring/gep.yaml
```diff
@@ -0,0 +1,17 @@
+title: SLO monitoring
+gep-number: 0041
+authors:
+- "@etiennnr"
+- "@xavier-re"
+- "@jguipi"
+- "@sonnyuplavan"
+steering: product
+steering-link: https://gardener.cloud/community/steering/technical|product/your-topic
```
rfranzke:
```suggestion
steering-link: https://gardener.cloud/community/steering/product/0041-slo-monitoring/
```

rfranzke (commented):
cc @istvanballok @chrkl 

For me, the main question is why this is proposed as an extension. Currently, all observability-related features live directly in `gardener/gardener`, even including the metering rules. What justifies moving this out now? What criteria or decision framework should guide whether something observability-related remains in-tree or becomes an extension? I'm having difficulty seeing a clear and consistent direction for monitoring-related functionality.

> Configurable SLOs

Do you have examples/use-cases for how/when users would use different values here?

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. GEP editors should help to ensure
+that the tone and content of the `Summary` section is useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+Although Gardener and Kubernetes expose a lot of metrics, there is presently no easy way to evaluate the availability of a shoot from the customer (shoot owner/operator) perspective. In the context that Gardener is meant to run a multitude of Kubernetes clusters at scale, operating it is usually not an easy endeavor. Having such means to measure and monitor the shoot's reliability is a key aspect from both customer happiness and operational excellence.
```
vlerenc: From a customer perspective: Maybe not perfectly, but there always was and is the "Kubernetes Control Plane Status" dashboard, which comes at least close.

Remark: I felt, that the perspective changed with the next sentences (now to operators). A pattern that will repeat and confuse during the rest of the GEP.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this GEP. Describe why the change is important and the benefits to users.
+-->
+
+Day-2 operation tooling is a must to help drive adoption. This also mimics industry best practices as described in the [SRE book](https://sre.google/sre-book/table-of-contents/).
```
vlerenc: If you want to place the SRE book link here, maybe you should first introduce day-2 and make the connect to SLO first?

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+
+This extension also enables Gardener operators and developers to make data-driven decisions about reliability vs feature prioritization while providing clear visibility into system health.
+
+This would benefit multiple user personas, including:
+
+- Operator: Having a ready to go extension to help with reliability and day-to-day operations
+- Developers: Having SLOs is a great tool to help prioritize between reliability and feature issues.
+- Users: Since SLOs are internal objectives meant for Gardener operators, they are normally not shared with the end users (shoot owners/operators). However, better reliability generally makes customers happy, which in turn drives user adoption.
```
vlerenc: Do you have a priority? At first you started with the users, now they are mentioned last. This repeats throughout the GEP and leaves a confused feeling whom you target first and foremost.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+-->
+- An SLO (service level objective) extension that provides a standardized framework for measuring and monitoring shoot cluster reliability, always from the customer perspective. By customer perspective, we want to focus on metrics that would reflect the customer's experience and satisfaction in operating their shoot clusters.
+
+### Non-Goals
+
+<!--
+What is out of scope for this GEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+- Provide SLAs to customers. SLOs are internal objectives for Gardener operators, and are not meant to be shared with customers (shoot owners/operators). However, such extension could be used to provide SLAs in the future, if desired.
```
vlerenc: I am confused. You mentioned end users twice above and now you write that the SLO this GEP is about is "not meant to be shared with customers".
Note: SLAs are a non-goal. That's clear. What is confusing is the rest that follows in the paragraph above.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### High level Architecture
+
+This extension delivers 4 core capabilities:
+
+- **Default and custom SLI**: A set of predefined SLIs/SLOs based on industry best practices for Kubernetes clusters and Gardener specifics, but always from the customer perspective. However, since the needs of each Gardener operator may vary, these defaults will be configurable. There should also be the ability to define custom SLIs based on specific metrics exposed by Gardener or the shoot clusters. However, those SLIs would need to be generic across the landscape.
```
vlerenc: Title says SLI, text says SLI and SLO. Clearly you want to define the SLIs, but the GEP and the explanation is also about the SLO, so I would mention them both consistently also in the title or else you confuse the reader (expecting SLI, then talking also about SLO).

What are the "default" / "predefined" SLIs/SLOs?
Note: Again different terms - more consistency in the wording would be beneficial to the reader, please.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+- **Configurable SLOs**: The SLO parameters (SLI, threshold, ...) should be configurable. Then a templating engine would generate the necessary Prometheus recording rules and Perses dashboards.
+- **SLO-based alerting**: Since we have the data to calculate SLO violations and burn rates (SRE best practice), we should also provide, as part of the extension, the capability to configure an Alertmanager based on those SLOs. Again, this should be configurable to fit the needs of each Gardener operator.
+- **Monitoring infrastructure**: The extension should provide the necessary monitoring infrastructure to collect, store, and visualize SLO-related metrics. This includes Prometheus rules for SLI calculation, Perses dashboards for visualization, Prometheus alerts for SLO violations, Alertmanager to manage those alerts, etc.
+
+The extension builds on the existing monitoring infrastructure (Prometheus operator, Perses operators, plutono annotations, ...), using a dedicated Prometheus instance in the runtime cluster to collect and aggregate SLO-specific metrics with minimal impact on the existing monitoring systems.
+
+![SLO extension high-level architecture](./slo-extension-plan.png)
```
vlerenc: It's not clear how "federation" against the network topology/direction will be implemented (push, VPN, something else). In Gardener, it is only guaranteed that the shoot can reach the seed and the seed can reach the garden cluster with a VPN between the seed and shoot cluster, but none between the garden and seed cluster. This requires some explanation before going on.

Consistency: The figure says Prometheus-SLO, the text below calls it `slo-prometheus`. Consistency helps the reader, please.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+-->
+### Initial SLOs
+
+The following SLOs are proposed as initial defaults, based on industry best practices for Kubernetes clusters and Gardener specifics, but always from the customer perspective. These are meant to be configurable, so they can be adapted to the needs of each Gardener operator. They are also meant to be adjusted on a regular basis based on real world data and experience, especially the thresholds. The goal is to have realistic and achievable SLOs that reflect the customer's experience and satisfaction in operating their shoot clusters.
+
+We decided to focus on 3 main topics for the initial SLOs:
+
+- Kube-apiserver is available, quick (latency) and without error.
+- Machines (nodes) are able to get created in a timely manner (join the cluster) and are running normally
+- Shoots are healthy, are able to get created and can reconcile in a timely manner
```
vlerenc: Please mention the technical metrics, if they already exist. That would make it more concrete. They already exist, I would think.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+  avg_over_time(
+    max(
+      probe_success{instance="https://api.internal_domain/healthz", type="seed"}
+      OR
+      probe_success{instance="https://kubernetes.default.svc.cluster.local/healthz", type="shoot"}
+      OR 
+      on() vector(0)
+    )[4w:5m]
+  ) * 100
+  ```
```
vlerenc: We already have the blackbox exporter shoot probe, which seems quite useful to me (goes through the internet, i.e. LB / front-door, I think). But it measures more than just the API server external availability. DWD does something similar, though I recall that it shortcuts on some infrastructures.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+> [!NOTE]
+> We are using a time windows of 28 days by default here, but this is going to be configurable based on the needs of each Gardener operator.
+
+> [!NOTE]
+> We are not aiming for perfection for the initial implementation, but rather to have a good starting point that can be improved over time based on real world data and experience. The goal is rather to have realistic and achievable SLOs that reflect the customer's experience and satisfaction in operating their shoot clusters. Hence, after the initial implementation, we should regularly review, adjust and add SLOs based on the data we collect and the feedback we get from customers and operators.
+
+### kube-apiserver general availability
+
+- SLI specification: The proportion of requests received with successful response from multiple sources vs the total amount of requests. In order to prevent false positive, a successful request is defined as at least 1 successful requests from any of the sources (inversely, a failing request means that it's failing from all sources are the same time). This is metric is processed at the shoot level.
+- SLI implementation:
```
ScheererJ: A different approach could be to check the istio metrics related to L7 loadbalancing if it is enabled instead of performing artificial probes. Unfortunately, the istio metrics are currently disabled.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+  (1-(sum(rate(apiserver_request_terminations_total[1h])) or on() vector(0) / sum(rate(apiserver_request_total[1h])))) * 100
+  ```
+
+- SLO Threshold: default TBD based on real world data, but this would be configurable
+Notes:
+  - N/A
+
+### Machine creation latency
+
+- SLI specification: The amount of machine trasitionning from `Pending` to `Running` within 20 minutes vs the total amount of nodes `Pending`in the last 20 minutes. If no nodes were pending in the last 20 minutes, the SLO default oto 100%. This is metric is processed at the shoot level.
```
ScheererJ: The 20 minute seem to relate to the default time out of `machine-controller-manager`. However, it can be configured individually per shoot cluster. Should this be considered here? For example, some shoot cluster owner may configure it more aggressively on infrastructures where node creation is fast. Then again, it may be necessary to increase this for bare metal machines, which take a lot longer to boot due to main memory checks.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+      ) == 0 // pending
+    ),
+    1
+  ) or on() vector(1)
+  ```
+
+- SLO Threshold: default TBD based on real world data, but this would be configurable
+- Notes:
+  - We need to implement a histogram metric that doesn't exist at the moment: `mcm_machine_creation_duration_minutes_bucket`
+  - confirm with MCM experts that the `Pending` state only happens during machine creation.
```
ScheererJ: A machine is also in `Pending` when the node-critical components are not ready, yet. From end-user perspective, this is still an unusable node, but it is not strictly related to `machine-controller-manager`.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+    )
+  )
+  ```
+
+  - `* on(name, namespace) group_left(node_name) mcm_machine_info `- Joins machine status with machine info to get the node_name label
+  - `label_replace(kube_node_created, "node_name", "$1", "node", "(.+)")` - Renames node label to node_name for matching
+
+- SLO Threshold: default TBD based on real world data, but this would be configurable
+- Notes:
+  - For now, we won’t take nodes less than 10 minutes old into account (default wait time for nodes to become ready is 20 minutes).
```
ScheererJ: Is the discrepancy (10 min vs. 20 min) desired? I understand this metrics rather as how nodes are available after they successfully joined the cluster.
WDYT?

ScheererJ (commented):
Thanks for creating the proposal.

timebertt (commented):
Thanks for opening this proposal.
My team has already set up basic availability monitoring/reporting for the shoot API servers and integrated it into STACKIT's internal availability monitoring for all other products. So I expect that there is potential for collaborating on this topic in general.
However, I'm skeptical whether this should actually be an extension and whether the suggested probes are technically feasible and product-wise sensible.
Most of my concerns were already mentioned inline by the other reviewers, so I refrained from duplicating them.

geps/0041-slo-monitoring/README.md
etiennnr: About the federation part, I explained the problem and possible alternatives in the `Notes/Constraints/Caveats` part.

I'll align the name for consistency indeed

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+
+## Alternatives
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+- **Reusing existing Prometheus instances**: Due to a high computational overhead to calculate SLOs, we didn't think it is a good idea to reuse the existing garden-prometheus, since this could potentially have an impact in other metrics. Also, since we need to keep the metrics for a longer period to calculate SLOs (more than the time window choosed for the SLOs), reusing the existing instances with extended retention could lead to resource contention. Hence, isolating the SLO ensures stability and separation of concerns.
+- **Implementing SLOs as part of Gardener core**: This functionnality also adds computational overhead, both for the shoot Prometheus and the aggregate Prometheus, so we believe this should be opt-in for Gardener operators that want to use it.
```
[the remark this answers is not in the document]
etiennnr: I'm not an expert on how much this prometheus is getting used, but this is actually something that was proposed/reviewed by monitoring colleagues. But the garden prometheus already federates metrics from seeds and from the garden-apiserver for every single shoot in it's landscape. So I'd guess quite busy already.

But, I'd say that data retention period is probably the main concern.

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+  ```promql
+  avg_over_time(
+    max(
+      probe_success{instance="https://api.internal_domain/healthz", type="seed"}
+      OR
+      probe_success{instance="https://kubernetes.default.svc.cluster.local/healthz", type="shoot"}
+      OR 
+      on() vector(0)
+    )[4w:5m]
+  ) * 100
```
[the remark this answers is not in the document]
etiennnr: As part of this extension, we won't scrape individual pods, only other prometheus servers (aka via federation). Hence, the metrics used here are already present. Just above, in the SLI specification, we say that this is at the shoot level. Hence, for this exemple, this entire block would be in the shoot prometheus as a recording rule (since the included metric are already scrapped there). Then the prometheus-slo would only federate the recording-rule result in the prometheus-shoot

geps/0041-slo-monitoring/README.md
etiennnr: Good point! However, isn't there a reason why those metrics are currently disabled ?

geps/0041-slo-monitoring/README.md
```diff
@@ -0,0 +1,358 @@
...
+
+  ```promql
+  avg_over_time(
+    max(
+      probe_success{instance="https://api.internal_domain/healthz", type="seed"}
+      OR
+      probe_success{instance="https://kubernetes.default.svc.cluster.local/healthz", type="shoot"}
+      OR 
+      on() vector(0)
+    )[4w:5m]
```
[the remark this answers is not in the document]
etiennnr: Well, it does work when I try it in plutono (of course we tested the ones that we could 😅) 
image

geps/0041-slo-monitoring/README.md
etiennnr: That seems quite obvious to me, but I guess not to everyone indeed!

geps/0041-slo-monitoring/README.md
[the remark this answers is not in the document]
timebertt: Interesting, I didn't know that this works. Thanks :)

timebertt:
Based on the discussion, should we close this @etiennnr?

etiennnr:
@timebertt should this be closed or instead merged as `rejected` since this is a proposal? I've just updated the `gep.yaml` file with `status: rejected`.

timebertt:
I'd prefer not to merge the document and to close this PR and the linked issue. 
Merging a rejected document can be confusing because the "rejected" status in the metadata is not visible within the document itself.
We should probably update the template/docs accordingly.

etiennnr:
All right then!

Issue #39, opened 2026-02-11
Live Control Plane Migration


rfranzke:
### Enhancement Description

- One-line enhancement description: Live Control Plane Migration
- Steering: Technical, https://gardener.cloud/community/steering/technical/2026-02-live-control-plane-migration/
- Gardener Enhancement Proposal: https://github.com/gardener/enhancements/tree/main/geps/0039-live-control-plane-migration
- Tracking Issue: 
- GEP (`g/enhancements`) update PR(s):

_Please keep this description up to date. This will help everybody to track the evolution of the enhancement efficiently._

Issue #28, opened 2026-02-11
Self-Hosted Shoot Clusters


rfranzke:
### Enhancement Description

- One-line enhancement description: Self-Hosted Shoot Clusters
- Steering: none
- Gardener Enhancement Proposal: https://github.com/gardener/enhancements/tree/main/geps/0028-self-hosted-shoot-clusters
- Tracking Issue: https://github.com/gardener/gardener/issues/2906
- GEP (`g/enhancements`) update PR(s):

_Please keep this description up to date. This will help everybody to track the evolution of the enhancement efficiently._

PR #52, opened 2026-02-26
GEP-0039: Applying feedback from the TSC review meeting

shafeeqes:
- How to categorize this PR:
/kind enhancement

- One-line PR description: Applying feedback from the TSC meeting on `09-02-2026`.

- Issue link: https://github.com/gardener/enhancements/issues/39

- Other comments: Please use the dedicated commits for easier review. The rest of the questions which could not be included in the GEP are answered in https://github.com/gardener/enhancements/pull/52#issuecomment-3965801520


rfranzke (commented):
>> 3 members per seed implies permanent quorum loss if seed-to-seed connectivity is lost.
>
> Answer -
>
> Yes, this is the case. The chances of this happening are very low. The risk of quorum loss also exists in a 5-member cluster and is actually lower than in a 6-member cluster. However, a major problem with a 5-member cluster is that Druid does not support scaling up from 1 → 2 → 3; it only supports scaling directly from 1 → 3 members. Therefore, for now, we will proceed with the 6-member cluster solution.

In the context of #28, druid will need to support scaling incrementally (1->2->3->more). Why can't we lever this and use this mechanism?

> In the case of Gardener, all seeds are upgraded at the same time.

Not really. This is very dependent on how the operators manage upgrades. The only automatism is that `gardenlet`s of `ManagedSeed`s get updated more or less at the same time (modulo jitters) when their parent `gardenlet` got upgraded.
And what do you mean with "In the case of Gardener"?

>> VPA recommendations:
>>
>> Not yet considered.
>> Open question: do VPACheckpoints need to be transferred as part of migration?
> Answer -
> 
> We had a discussion with the OG CPM team.

What is the "OG CPM team"? Generally, the handling for VPACheckpoints is worth a sentence in the document.

> Currently, there is no planned solution, and this issue should be handled separately from this GEP.

Because VPACheckpoints are also not handled for regular control plane migration, or why?

geps/0039-live-control-plane-migration/README.md
```diff
@@ -141,77 +105,94 @@ To trigger **Live CPM**, a new operation annotation, `gardener.cloud/operation=l
...
+      reason: RestoringExtensions
+      message: "Restoring extension states on destination seed"
 ```
 
+Each gardenlet (source and destination) updates relevant conditions as migration progresses through different steps.
+
 #### Lease Management
 
-Leases are used to coordinate ownership and leadership. Components that manage resources in the Shoot cluster and are deployed in the Seed cluster should maintain their leases in the Shoot cluster. This ensures that only one control plane instance (source or destination) performs operations on the resources at any given time, preventing split-brain scenarios.
+Leases are used to coordinate ownership and leadership. Components that manage resources in the `Shoot` cluster and are deployed in the `Seed` cluster should maintain their leases in the `Shoot` cluster. This ensures that only one control plane instance (source or destination) performs operations on the resources at any given time, preventing split-brain scenarios.
```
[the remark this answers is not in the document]
shafeeqes: For extensions, right. But we still want to deploy control plane components such as `kube-controller-manager` and `kube-scheduler` on the destination while they are running in the source.

shafeeqes:
> In the context of https://github.com/gardener/enhancements/issues/28, druid will need to support scaling incrementally (1->2->3->more). Why can't we lever this and use this mechanism?

We will get back to you on this after discussing with the etcd-druid team (/cc @CaptainIRS @shreyas-s-rao @unmarshall).

> Not really. This is very dependent on how the operators manage upgrades.

Agree.

> Generally, the handling for VPACheckpoints is worth a sentence in the document.

Will add it.

> Because VPACheckpoints are also not handled for regular control plane migration, or why?

Yes, they are not handled currently.

ScheererJ (commented):
Thanks for updating the proposal.

CaptainIRS (commented):
>> Druid does not support scaling up from 1 → 2 → 3; it only supports scaling directly from 1 → 3 members. Therefore, for now, we will proceed with the 6-member cluster solution.

> In the context of https://github.com/gardener/enhancements/issues/28, druid will need to support scaling incrementally (1->2->3->more). Why can't we lever this and use this mechanism?

etcd-druid supports incremental scale-up of 1->2->3->n in the existing codebase. I believe the docs show the 1->3 use-case just because in gardener's context, a non-HA to HA transition is a scale-up of 1->3.

geps/0039-live-control-plane-migration/README.md
```diff
@@ -104,7 +105,9 @@ To trigger **Live CPM**, a new operation annotation, `gardener.cloud/operation=l
...
+- The source and destination seed clusters must not be in Distant Regions. Gardener can use the `garden-scheduler` [ConfigMap](https://github.com/gardener/gardener/blob/v1.134.1/docs/concepts/scheduler.md#minimal-distance-strategy) to evaluate the inter-seed "distance" (for example, network latency).
+  - Operators can define a landscape-specific threshold by annotating this ConfigMap with `migration.gardener.cloud/inter-seed-distance-threshold=<value>` (for example, `180` when the distance metric is network latency in milliseconds). The value of `180` was derived from extensive testing across different cloud providers. If operators use a different distance metric, they must adjust the threshold accordingly.
+  - If the scheduler ConfigMap or the threshold annotation is not provided, Gardener cannot determine the distance between seeds. In such cases, operators may still force a migration by annotating the Shoot with `migration.gardener.cloud/allow-distant-regions=true`, fully aware of the associated risks.
```
timebertt: In simpler single-provider/region setups (like ours), there is no such ConfigMap, as all seeds are running on the same infrastructure and within the same region. We should probably just allow the live migration if the ConfigMap doesn't exist.

timebertt (commented):
Thanks for updating the proposal document. 
I'm still not convinced by the 6-member etcd cluster approach. In my opinion, it is too high a risk. We should further look into a 5-member etcd with 1-member surge/unavailable (member-by-member migration).

geps/0039-live-control-plane-migration/README.md
ary1992: We need the annotation to make it explicit about the risk associated with migration for the cases where migration is done across a distant region, or migration done without a configmap (which has latency information across different regions).

We can allow migration without configmap, if the migration is performed in the same region.

timebertt:
Sorry, but I lost track of the current state of this PR.
Is it ready for another round of in-depth review? I.e., have you addressed all feedback?

timebertt (commented):
Thanks for updating the GEP. 

While I think it's in better shape now, I'm still a bit "scared" by the migration flow's complexity (see inline comments). I would have hoped that we can achieve the goal with a much simpler flow like this:
1. expose the source etcd members and prepare components before API server for migration (Infrastructure, some Extensions, BackupEntry, etc.)
2. join 2 destination etcd members and deploy components and API server
3. make the leader switch: move one etcd member from source to seed, migrate the DNS record
4. migrate the remaining components after API server (gardener-resource-manager, Worker, Network, other Extensions, etc.)
5. cleanup everything on the source side
6. cleanup everything on the destination seed and remove the temporary VPN shoot client

This flow would (kind of) follow the usual reconciliation flow and would be more intuitive than what you're outlining here.

shafeeqes:
@rfranzke @timebertt @ScheererJ, Could you please take another look?

rfranzke (commented):
/lgtm

timebertt:
Very nice, thanks for updating the GEP. The simplified flow looks much better and easier to follow to me.
I'd be happy to schedule the next steering session to finalize this proposal and get it merged :)

rfranzke:
Recording and meeting minutes for second TSC round have been posted: https://gardener.cloud/community/steering/0039-live-control-plane-migration/ (see "Round 2").

Please incorporate the remaining changes. :)

shafeeqes:
Please take another look @rfranzke @timebertt!

geps/0039-live-control-plane-migration/livecpm-flow.svg
timebertt, line 1: I'm missing the update addressing this decision:

> 5-member cluster: join 3 on destination, briefly scale to 6, immediately remove one source member (back to 5), then remove remaining two. Minimizes the 6-member window.

geps/0039-live-control-plane-migration/README.md
```diff
@@ -141,76 +103,94 @@ To trigger **Live CPM**, a new operation annotation, `gardener.cloud/operation=l
...
 
 #### Member removal from the cluster
-- During the six-member cluster formation, the destination seed cluster’s etcd CR contains fields to bootstrap from the source cluster. To complete the migraion, the etcd member count needs to be brought back to three by removing the members that are part of the source seed cluster. For this, the bootstrap with source cluster field is removed from the destination cluster’s etcd CR. At the time of writing the GEP, it was decided that druid will use [EtcdOpsTask](https://github.com/gardener/etcd-druid/issues/1047) to remove members, and a new member removal task will be introduced.
+
+- During the five-member cluster formation, the destination etcd CR holds the bootstrap fields needed to join the source cluster. To finish the migration, one more member is added to the destination etcd CR (bringing the total to six) as mentioned before, and then the three source members are removed by clearing those bootstrap fields. Member removal is done using the [GEP-28](../0028-self-hosted-shoot-clusters/) mechanism, which calls HTTP endpoints on the backup-restore sidecar.
```
timebertt: > Reuse GEP-28 member removal: destination-side etcd-druid orchestrates removal via backup-restore HTTP endpoint. Leadership transfer before removing the leader. No EtcdOpsTask.

From reading the document, it's not clear to me that the destination side triggers/orchestrates the removal of source etcd members. Can you clarify who calls the HTTP endpoints of etcd-backup-restore?

timebertt (commented):
Thanks for updating the proposal. I'm still missing a few changes based on one decision we made last week.

geps/0039-live-control-plane-migration/README.md
shafeeqes: >  one more member is added to the destination etcd CR (...) and then the three source members are removed by clearing those bootstrap fields.

The bootstrap fields are only present on the destination etcd CR.
Should I mention that the destination etcd-druid does it explicitly?

geps/0039-live-control-plane-migration/livecpm-flow.svg
shafeeqes: We can only make destination 3-source 2, after the KAPI has started on the destination and the DNS record is switched. At that point, we are anyway removing all the source members at once. So the mentioned approach doesn't give us many benefits.

timebertt: > We can only make destination 3-source 2, after the KAPI has started on the destination and the DNS record is switched. 

Why is that?

geps/0039-live-control-plane-migration/README.md
timebertt: Got it. Yeah, that would improve the readability

geps/0039-live-control-plane-migration/livecpm-flow.svg
shafeeqes: Otherwise, if the network connection is lost, the DNS record is still pointing to the source, but that KAPI cannot serve any requests anyway with the 2 member etcd cluster

timebertt: Ah, good point. Then let's keep it like it is

geps/0039-live-control-plane-migration/README.md
shafeeqes: Done.

geps/0039-live-control-plane-migration/README.md
```diff
@@ -156,7 +156,7 @@ For both source and destination seeds, the following resources are created:
 
 - A single `Gateway` resource.
 - A `VirtualService` per etcd member to perform host-based routing to the respective etcd member `Service`, and a `DestinationRule` per etcd member to configure traffic policies for those connections.
-- One `DNSRecord` per etcd member, each pointing to the `Seed` Istio `IngressGateway` LoadBalancer, with traffic routed to the correct member via the corresponding `VirtualService`.
+- One `DNSRecord` per etcd member ( With domain name <etcd-pod-name>.<etcd-pod-namespace>.<seed-name>.<default-domain>), each pointing to the `Seed` Istio `IngressGateway` loadbalancer, with traffic routed to the correct member via the corresponding `VirtualService`.
```
rfranzke: Nit: I think it's better to use the internal domain since not all `Shoot`s might have an external domain and the Gardener system as a whole might not have default domains.

rfranzke (commented):
Can you also talk about

> Exact gardenlet version match (incl. patch) enforced before and during migration. API server may additionally reject mismatched versions.

?

rfranzke (commented):
/lgtm

timebertt (commented):
No more comments from my side, this looks good!
Thanks for addressing all of my concerns :) 
/lgtm

PR #53, opened 2026-02-26
GEP-0054: Worker Capability Requirements for Machine Image Selection

Roncossek:
- How to categorize this PR:
/area os usability
/kind enhancement

- One-line PR description: Adds new GEP 0054 - Worker Capabilities for Machine Image Selection

- Issue link: https://github.com/gardener/enhancements/issues/54

- Other comments:


geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+  - [Drawbacks](#drawbacks)
+
+
+## Summary
+
+This GEP extends [GEP-0033 (Machine Image Capabilities)](../0033-machine-image-capabilities/README.md) to support selecting machine images based on worker-defined properties in addition to machine type characteristics. While GEP-0033 enables image selection based on hardware capabilities of machine types (CPU architecture, hypervisor type, etc.), this proposal introduces worker capabilities that describe software/feature requirements requested by the worker configuration. Both machine type capabilities and worker capabilities must be satisfied for image selection to succeed.
+
+## Motivation
+
+Currently, machine image capabilities as defined in GEP-0033 only support selecting images based on hardware properties of the machine type (CPU architecture, hypervisor type, bare-metal vs virtualized). However, users increasingly need the ability to select images based on software or feature requirements that are specified in the worker configuration, such as:
```
rfranzke: Isn't the real motivation the defaulting? Users (already today) know what their worker pools will need. If they need in-place updates, they can look up a proper machine image/type in the `CloudProfile`. Same goes for other capability requirements.

What you seem to want to achieve is that they don't have to look this up and manage themselves, right?

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+    if worker.Machine.GPU != nil {
+        switch worker.Machine.GPU.Type {
+        case "nvidia":
+            capabilities[CapabilityNameGpuSupport] = []string{CapabilityValueGpuSupportNvidia}
+        case "amd":
+            capabilities[CapabilityNameGpuSupport] = []string{CapabilityValueGpuSupportAmd}
+        case "intel":
+            capabilities[CapabilityNameGpuSupport] = []string{CapabilityValueGpuSupportIntel}
+        default:
+            capabilities[CapabilityNameGpuSupport] = []string{CapabilityValueGpuSupportNone}
```
rfranzke: We even go beyond and not only hard-code capability names, but also concrete values (like `nvidia`, etc.)? Then I don't understand the whole effort that was put in GEP-33 to make the API as generic as possible.

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+The key distinction between capability types:
+- **Machine type capabilities**: Describe what the hardware supports (e.g., ARM64 architecture, virtualized hypervisor)
+- **Worker capabilities**: Describe what features the worker configuration requests (e.g., secure boot enabled, in-place updates required)
+
+Worker capabilities are derived from worker specification properties. To prevent naming conflicts between Gardener-reserved capabilities and custom or provider-defined capabilities, reserved capability names use a `gardener-` prefix.
+
+### Notes/Constraints/Caveats
+
+- Worker capabilities are **additive** to the existing capability system. Existing CloudProfiles and image selection continue to work without modification.
+- The `gardener-` prefix is reserved for Gardener core capabilities. Provider extensions should use their own prefixes (e.g., `azure-`, `aws-`) for provider-specific capabilities.
```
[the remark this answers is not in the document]
ScheererJ: Why not just combine the prefix, e.g. `gardener-azure-` and `gardener-aws-`? This way users can use anything, but the prefix `gardener-`.

ScheererJ (commented):
Thanks for the proposal.

From my perspective, it should be adapted to be less abstract. I wondered most of the time what this is actually about until I reached the design details. It should be possible to make this more tangible and understandable by adding examples earlier and using a better name as Rafael suggested.

Roncossek:
Thanks for the valuable feedback so far! I've reworked the GEP substantially.

I structured the proposal around the two currently known use cases that connect worker pool fields with GEP-33 capabilities:
- in-place node updates as a worker update strategy
-  secure boot support (planned)

This gives the proposal more direction and makes the benefits clearer.

I also narrowed the scope: the GEP now only covers capabilities owned by Gardener under the gardener- prefix. Provider-extension-owned capabilities driven by WorkerConfig were deferred to a future GEP once/if actual use cases exist that would justify such changes.

Would appreciate another look.
cc @ScheererJ @rfranzke @vpnachev

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+This GEP reserves the prefix `gardener-` inside GEP-0033 capabilities for this family of features and offers an option for existing or future typed worker pool fields (e.g. `updateStrategy`) into the GEP-0033 selection algorithm. **Users keep configuring features via typed worker fields**; Gardener internally derives the matching capability requirements.
+
+## Motivation
+
+Today, when a user enables in-place updates on a worker pool, nothing checks that the selected machine image and machine type actually support them. The misconfiguration surfaces only at runtime. Avoiding it requires manually finding a machine-type / machine-image pair in the CloudProfile that is compatible with the worker pool configuration, and keeping that selection valid across maintenance updates. This means for now automatic OS update operations are off the table for in-place update pools.
```
ScheererJ: I am not sure I can follow the conclusion here. Could you please explain why you arrive at the last sentence?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,166 @@
...
+**Key terms:**
+
+- **Reserved capability** — capability whose name starts with `gardener-`; key and values defined by Gardener.
+- **Capability requirement** — `(name, value)` pair derived internally from a worker pool's typed fields. Never written by users.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|---|---|
+| An operator accidentally registers a `gardener-*` capability with values that don't match Gardener's authoritative definition | CloudProfile admission rejects the registration |
```
ScheererJ, line 113: What will happen if an operator has already added a `gardener-*` capability now. Admission cannot reject it now. What will happen once this is implemented?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,166 @@
...
+| Reserved-capability definitions drift between a Gardener release and a CloudProfile maintained by an operator | CloudProfile admission validates registered values against the Gardener version in use; mismatches surface immediately rather than at runtime |
```
ScheererJ, line 114: I am not sure I understand this. Admission only validates on changes. If the `CloudProfile` is never changed, but the gardener-internal values change admission will never be called. How is such drift mitigated?

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,166 @@
...
+GEP-0033 introduced a selection algorithm based on capability compatibility:
+
+```go
+AreCapabilitiesCompatible(imageFlavor, machineType, capabilityDefinitions)
+```
+
+This GEP adds a third input:
+
+```go
+AreCapabilitiesCompatible(imageFlavor, machineType, workerRequirements, capabilityDefinitions)
```
ScheererJ: Not sure if you are counting zero-based here, but I would argue it adds the fourth input to the function.

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+The same algorithm is invoked from three existing call sites:
+
+1. [Shoot validator admission](https://github.com/gardener/gardener/blob/e6263d6a575e4181f0289345803ccb59117605f6/plugin/pkg/shoot/validator/admission.go#L1000) — rejects user-selected machine image / machine type pairs that are incompatible with the worker pool's capability requirements.
+2. [Shoot mutator admission](https://github.com/gardener/gardener/blob/d9897865ab9181c307efdfa93f14268fcd09fe88/plugin/pkg/shoot/mutator/admission.go#L559) — picks a default machine image compatible with the chosen machine type and worker pool when the user does not specify one.
+3. [Maintenance controller](https://github.com/gardener/gardener/blob/d9897865ab9181c307efdfa93f14268fcd09fe88/pkg/controllermanager/controller/shoot/maintenance/helper/helper.go#L27) — chooses a machine image compatible with the worker pool's machine type and capability requirements during automatic updates.
+4. Worker controller in provider extensions.
```
ScheererJ: Again, I am wondering why the document says `three` while I count four. Pleas clarify.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+| Concern | Validated by |
+|---|---|
+| `gardener-*` capability registrations match Gardener's authoritative list | gardener-apiserver CloudProfile admission |
+| Worker pool's capability requirements satisfied by the selected machine image and machine type | gardener-apiserver shoot admission |
+| Maintenance image selection respects all requirements | maintenance controller |
+
+## Drawbacks
+
+- **Internal complexity grows.** Selection gains a third input and a derivation step. The user-side simplification is the explicit trade.
```
ScheererJ: Does this refer to the function with four input parameters?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,166 @@
...
+- **Operator setup per CloudProfile.** Operators who want a reserved feature must register the capability in `spec.machineCapabilities` and declare per-image-flavor support. CloudProfiles that expose no reserved features are unaffected.
+- **Coupling between Gardener releases and CloudProfile content.** Reserved definitions are owned by Gardener; operators must keep CloudProfiles in sync with the version they run. CloudProfile admission catches mismatches.
```
ScheererJ: Admission can only catch something when the `CloudProfile` is actually changed. Do we always expect this?

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Drawbacks
+
+- **Internal complexity grows.** Selection gains a third input and a derivation step. The user-side simplification is the explicit trade.
+- **Operator setup per CloudProfile.** Operators who want a reserved feature must register the capability in `spec.machineCapabilities` and declare per-image-flavor support. CloudProfiles that expose no reserved features are unaffected.
+- **Coupling between Gardener releases and CloudProfile content.** Reserved definitions are owned by Gardener; operators must keep CloudProfiles in sync with the version they run. CloudProfile admission catches mismatches.
+
+## Alternatives
+
+- **Generic `capabilities` map on the worker pool API.** Most consistent with GEP-0033, but rejected for three reasons: (1) it forces users to learn the CloudProfile's capability vocabulary to configure standard features; (2) most capabilities are infrastructure-level concerns that are irrelevant to shoot users (e.g. hypervisor type, or which bare-metal machine type works with which image) — exposing them on the worker pool API would surface implementation detail with no user benefit; (3) it leaks an internal contract into operator-facing API.
```
ScheererJ: Please structure your reasons in a way that it is also easy to digest and not a wall of text.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,226 @@
...
+| Provider extensions not updated | Validation of the shoot by the extension is only required if the provider supports custom worker capabilities |
+
+## Design Details
+
+### Reserved Capability Names
+
+The following example capability names are reserved for Gardener core use. All reserved capabilities use the `gardener-` prefix:
+
+```go
+const (
```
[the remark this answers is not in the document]
Roncossek: I updated the capability name and values to be kebab case. As they are not purely API fields with an enum but more like keys and labels in the cloudprofile imo. Here kebab case would be allowed if i read the documentation correctly:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants,

But im open to change reserved capability names and values to CamelCase aswell.

Its a different story for the worker Api here CamelCase must be used by each feature. (as in-place already does)

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+**Example Locations:**
+1. **Shoot Validator Admission** ([worker admission](https://github.com/gardener/gardener/blob/e6263d6a575e4181f0289345803ccb59117605f6/plugin/pkg/shoot/validator/admission.go#L1000))
+
+2. **Shoot Mutator Admission** ([default image selection](https://github.com/gardener/gardener/blob/d9897865ab9181c307efdfa93f14268fcd09fe88/plugin/pkg/shoot/mutator/admission.go#L559))
+
+3. **Maintenance Controller** ([image selection helper](https://github.com/gardener/gardener/blob/d9897865ab9181c307efdfa93f14268fcd09fe88/pkg/controllermanager/controller/shoot/maintenance/helper/helper.go#L27C1-L27C149))
+
+### Provider-Specific Worker Properties
+
+Provider extensions can define additional worker properties that map to capabilities. This is achieved through the existing `WorkerConfig` extension mechanism:
```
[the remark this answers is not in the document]
Roncossek: I put my reasoning against capability maps in the worker it in the #Alternatives section and dropped the provider specific features:

(1) it forces users to learn the CloudProfile's capability vocabulary to configure worker features; 
(2) most capabilities are infrastructure-level concerns that are irrelevant to shoot users (e.g. hypervisor type, or which bare-metal machine type works with which image). Exposing them on the worker pool API would surface implementation detail with no user benefit; 
(3) it leaks an CloudProfile internal contract into user-facing API.
(4) changing/removing capabilities today can be done isolated within a CloudProfile. This would add a dependency to the workers using that CloudProfile.

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Proposal
+
+Extend the capability matching system from GEP-0033 to include a third dimension: worker capabilities. The image selection algorithm currently performs a capability intersection between machine image capabilities and machine type capabilities. This proposal adds worker capabilities to the intersection, ensuring selected images satisfy all three capability sets.
+
+The key distinction between capability types:
+- **Machine type capabilities**: Describe what the hardware supports (e.g., ARM64 architecture, virtualized hypervisor)
+- **Worker capabilities**: Describe what features the worker configuration requests (e.g., secure boot enabled, in-place updates required)
+
+Worker capabilities are derived from worker specification properties. To prevent naming conflicts between Gardener-reserved capabilities and custom or provider-defined capabilities, reserved capability names use a `gardener-` prefix.
```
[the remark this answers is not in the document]
Roncossek: I restructured the motivation around the existing use case of in place node updates. I hope that makes it clearer. 

NamespacedCloudprofiles are unrelated.

geps/0040-worker-capabilities/README.md
Roncossek: Thanks for catching that. I was a bit too focused on the case that multiple flavors exist in a version, and ignored reality... 

Automatic in-place node updates are possible today. As long as all image references used within that version support that feature. 

Looking at the [2150.4.0 Release](https://github.com/gardenlinux/gardenlinux/releases/tag/2150.4.0) of a Gardenlinux Version: we can see one version publishes multiple flavors. One of this flavors (the USI) supports in place updates. 

If we include that flavor into the same version as the other ones, Gardener currently lacks the mechanism to select this flavor within the version. As no mapping from the worker to the capabilityFlavor can be done.

```
  machineImages:
    - name: gardenlinux
      versions:
        - version: "2150.4.0"
           inPlaceUpdates:                                                                                                                      │
             supported: true
          capabilityFlavors:
            - architecture: [amd64]
              gardener-update-type: [rolling]
             # the mechanism to ensure the flavor below will be selected is missing 
            - architecture: [amd64]
              gardener-update-type: [in-place]
```

What works is of course have a version that only includes the in place update flavor. Then automatic in-place updates work with some custom filter logic in the maintenance controller on version level (filter versions that dont have `inPlaceUpdates.supported: true`.
 
```
  machineImages:
    - name: gardenlinux
      versions:
        - version: "2150.4.0"
          capabilityFlavors:
            - architecture: [amd64]
          
     
        - version: "2150.4.0-inplace"
           inPlaceUpdates: 
             supported: true
          capabilityFlavors:
            - architecture: [amd64]


```

Roncossek: I updated the motivaton to be correct and explain the issue in a bit more detail.

geps/0040-worker-capabilities/README.md
Roncossek: I added some more comments. explaining that now 3 capability structs will be passed and the capabilityDefinitions (which i did not count)

geps/0040-worker-capabilities/README.md
Roncossek: Corrected the counting 🙈

geps/0054-worker-capability-requirement/README.md
Roncossek: You are right: drift is only caught when the `CloudProfile` is updated.

If the feature is not defined within a worker, nothing changes. 
If it is used, `deriveWorkerCapabilityRequirements` should only map the worker field to a reserved capability when the `CloudProfile` defines that capability with the expected values and error otherwise. Here I see two options:
- create the worker without that feature
- reject the worker as a feature is selected that not supported by the Cloudprofile.

I lean into the second option. What do you think?

Once the `CloudProfile` is updated, admission will reject invalid `gardener-*` values. Since `spec.machineCapabilities` is internal `CloudProfile` data, fixing it should be manageable with a rename or value update.

The hard case is when an operator already uses the same capability name and values for a different meaning. That is difficult to prevent. But also very unlikely that the same key values are used for a different purpose.

geps/0054-worker-capability-requirement/README.md
Roncossek: I guess the question overlaps with the mitigation one. I wrote an update there.

> I am not sure I understand this. Admission only validates on changes. If the `CloudProfile` is never changed, but the gardener-internal values change admission will never be called. How is such drift mitigated?

geps/0040-worker-capabilities/README.md
Roncossek: I updated the Alternatives part, hope its easier to digest now.

geps/0040-worker-capabilities/README.md
Roncossek: I updated the GEP and especially the motivation, I guess that should explain better why some capabilities need gardener ownership. Or be outside of the capability system all together.

geps/0040-worker-capabilities/README.md
Roncossek: I updated the motivation and hope this one has clearer reasoning. 🙈

geps/0054-worker-capability-requirement/README.md
```diff
@@ -123,21 +128,45 @@ Only typed worker pool fields drive requirements — users of the shoot resource
...
-AreCapabilitiesCompatible(imageFlavor, machineType, capabilityDefinitions)
+AreCapabilitiesCompatible(imageFlavor, machineType, capabilityDefinitions){
+	defaultedCapabilities1 := GetCapabilitiesWithAppliedDefaults(imageFlavor, capabilityDefinitions)
+	defaultedCapabilities2 := GetCapabilitiesWithAppliedDefaults(machineType, capabilityDefinitions)
+
+	commonCapabilities := GetCapabilitiesIntersection(defaultedCapabilities1, defaultedCapabilities2)
+	// If the intersection has at least one value for each capability, the capabilities are compatible.
+	for _, values := range commonCapabilities {
+		if len(values) == 0 {
+      return false
```
ScheererJ, line 185: Seems like this line uses spaces instead of tabs.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -123,21 +128,45 @@ Only typed worker pool fields drive requirements — users of the shoot resource
...
 
-This GEP adds a third input:
+This GEP adds a third capability input for the worker pool's derived requirements:
 
 ```go
-AreCapabilitiesCompatible(imageFlavor, machineType, workerRequirements, capabilityDefinitions)
+AreCapabilitiesCompatible(imageFlavor, machineType, workerRequirements, capabilityDefinitions){
+	defaultedCapabilities1 := GetCapabilitiesWithAppliedDefaults(imageFlavor, capabilityDefinitions)
+	defaultedCapabilities2 := GetCapabilitiesWithAppliedDefaults(machineType, capabilityDefinitions)
+  defaultedCapabilities3 := GetCapabilitiesWithAppliedDefaults(workerRequirements, capabilityDefinitions)
```
ScheererJ, line 198: This line also uses spaces instead of tabs.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -123,21 +128,45 @@ Only typed worker pool fields drive requirements — users of the shoot resource
...
+
+	commonCapabilities := GetCapabilitiesIntersection(defaultedCapabilities1, defaultedCapabilities2, defaultedCapabilities3)
+	for _, values := range commonCapabilities {
+		if len(values) == 0 {
+      return false
```
ScheererJ, line 203: This line also uses spaces instead of tabs.

geps/0054-worker-capability-requirement/README.md
ScheererJ: I am not sure if I understand it to the fullest extend, but your reasoning sounds good.

geps/0054-worker-capability-requirement/README.md
Roncossek: One cloud reserve the `gardener-*` namespace and only after some releases (3 or 4) allow adding the first reserved capability. Then I guess we could assume that one Cloudprofile update was triggered?

geps/0054-worker-capability-requirement/README.md
Roncossek: but in general, we should separate the namespace reservation from the first feature using that namespace. We could first reserve the namespace and after several releases add the first reserved capability. After that time we could assume (still with a breaking change release note) that all cloudprofiles were once updated.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,201 @@
...
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Design Details](#design-details)
+    - [Reserved Capability Names](#reserved-capability-names)
+    - [Deriving Capability Requirements](#deriving-capability-requirements)
+    - [Image Selection Algorithm](#image-selection-algorithm)
+    - [Validation](#validation)
+  - [Drawbacks](#drawbacks)
+  - [Alternatives](#alternatives)
```
rfranzke: This section is no longer part of the template.

geps/0054-worker-capability-requirement/gep.yaml
```diff
@@ -0,0 +1,11 @@
+title: Worker Capabilities for Machine Image Selection
```
rfranzke, line 1: Why is this in `geps/0040` while the README is in `0054`?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,201 @@
...
+1. **machine type** metadata is not defined
+1. compatibility is enforced by several separate **filter implementations**: e.g. [maintenance operations](https://github.com/gardener/gardener/blob/b8a336572b2305befa05d358480ff22f3e11d0a9/pkg/controllermanager/controller/shoot/maintenance/helper/helper.go#L206), [shoot admission](https://github.com/gardener/gardener/blob/159fbd0df8a3a272bfdf054638ce1ec1f45ccc9e/plugin/pkg/shoot/validator/admission.go#L1594-L1610), [default image selection](https://github.com/gardener/gardener/blob/159fbd0df8a3a272bfdf054638ce1ec1f45ccc9e/plugin/pkg/shoot/mutator/admission.go#L632-L634), ...
+1. **image version flavor selection** is missing; if more than one flavor is grouped under a machine image version, selecting a compatible flavor is not guaranteed
+
+As a result, Gardener can accept or derive combinations whose full compatibility is not expressed in one place, and operators may need to split images into separate versions just to encode flavor-specific support and allow automatic image maintenance to work.
```
rfranzke, line 22: I think a more concrete configuration example would help here to outline the current problems.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+The underlying problem: these features need a shared compatibility check across worker pool configuration, machine image flavors, and machine types. This GEP therefore builds on the existing GEP-33 capability mechanism by reserving a Gardener-owned capability namespace and by deriving capability requirements from typed worker pool fields, so that admission and image selection can consistently reject incompatible combinations. Once implemented, only minimal changes are required to the worker pool API to support new features.
```
rfranzke: only minimal?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,201 @@
...
+const (
+    // gardener-update-type — drives in-place node updates (GEP-0031).
+    GardenerCapabilityUpdateType        = "gardener-update-type"
+    GardenerCapabilityUpdateTypeRolling = "rolling"
+    GardenerCapabilityUpdateTypeInPlace = "in-place"
+
+    // gardener-boot-type — drives secure boot (planned).
+    GardenerCapabilityBootType         = "gardener-boot-type"
+    GardenerCapabilityBootTypeStandard = "standard"
+    GardenerCapabilityBootTypeSecure   = "secure"
```
rfranzke, line 133: If Gardener needs to know/be aware of certain capabilities for proper orchestration, wouldn't it perhaps be cleaner to augment the `MachineCapabilities` API and make those first-class citizens in the API? The generic mechanism is great for properties that don't require Gardener knowledge, but using the proposed prefix approach comes with its own problems ([ref](https://github.com/gardener/enhancements/pull/53/changes#diff-6619dfc0f72182e15698f6555a4a7920fea986a277d6e83f91cd6abcf19e553bR71)).

I understand that this might be easier/faster to implement, but API/usage-wise we could do better with dedicated fields in the API for Gardener-relevant capabilities.

It could internally still be encoded like the generic capabilities such that the matching algorithms work as they are.

WDYT?

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,201 @@
...
+|---|---|
+| `gardener-*` capability registrations match Gardener's authoritative list | gardener-apiserver CloudProfile admission |
+| Worker pool's capability requirements satisfied by the selected machine image and machine type | gardener-apiserver shoot admission |
+| Maintenance image selection respects all requirements | maintenance controller |
+
+## Drawbacks
+
+- **Internal complexity grows.** Selection gains a third input and a derivation step. The user-side validation together with a framework for future feature development is the explicit trade.
+- **Operator setup per CloudProfile.** Operators who want a reserved feature must register the capability in `spec.machineCapabilities` and declare per-image-flavor support. CloudProfiles that expose no reserved features are unaffected.
+- **Coupling between Gardener releases and CloudProfile content.** Reserved definitions are owned by Gardener; operators must keep CloudProfiles in sync with the version they run to use new features.
```
rfranzke, line 231: x-ref above comment

geps/0040-worker-capabilities/README.md
```diff
@@ -0,0 +1,226 @@
...
+    - [Worker Properties to Capabilities Mapping](#worker-properties-to-capabilities-mapping)
+    - [Updated Image Selection Algorithm](#updated-image-selection-algorithm)
+    - [Provider-Specific Worker Properties](#provider-specific-worker-properties)
+    - [Component Changes](#component-changes)
+  - [Drawbacks](#drawbacks)
+
+
+## Summary
+
+This GEP extends [GEP-0033 (Machine Image Capabilities)](../0033-machine-image-capabilities/README.md) to support selecting machine images based on worker-defined properties in addition to machine type characteristics. While GEP-0033 enables image selection based on hardware capabilities of machine types (CPU architecture, hypervisor type, etc.), this proposal introduces worker capabilities that describe software/feature requirements requested by the worker configuration. Both machine type capabilities and worker capabilities must be satisfied for image selection to succeed.
```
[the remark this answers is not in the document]
Roncossek: What about "Worker Capability Requirements"?

geps/0054-worker-capability-requirement/README.md
Roncossek: removed section

geps/0054-worker-capability-requirement/gep.yaml
Roncossek: whooops, all in `0054` now

geps/0054-worker-capability-requirement/README.md
Roncossek: do you mean something like this?

```
  machineImages:
    - name: gardenlinux
      versions:
        - version: "2150.4.0"
          capabilityFlavors:
            - architecture: [amd64]
          
        - version: "2150.4.0-inplace"
           inPlaceUpdates: 
             supported: true
          capabilityFlavors:
            - architecture: [amd64]
```

geps/0054-worker-capability-requirement/README.md
Roncossek: Just some field in the worker pool API and the mapping for deriving the capabilities. 

Then the admission and maintenance part is done. (only the cloudprofile must be updated by the operators)

Then of course the actual feature must still be implemented.

geps/0054-worker-capability-requirement/README.md
rfranzke: Yes, some YAMLs that demonstrate the current problems and reinforce the need for a change (this proposal).

geps/0054-worker-capability-requirement/README.md
Roncossek: you mean dedicated fields in the Cloudprofile API? 

```yaml
# current state
spec:
  machineCapabilities:
    - name: architecture
      values: ["amd64", "arm64"]
    ...
```

like this: 
```yaml
spec:
  gardenerOSCapabilities:
    nodeUpdateTypes: ["rolling", "in-place"]
    bootTypes: ["secure", "legacy"]
    ...
  machineCapabilities:
    - name: architecture
      values: ["amd64", "arm64"]
    ...
```

or like that:
```yaml
spec:
  machineCapabilities:
    gardenerOwned:
      nodeUpdateTypes: ["rolling", "in-place"]
      bootTypes: ["secure", "legacy"]
      ...
    custom:
      - name: architecture
        values: ["amd64", "arm64"]
      ...
```

geps/0054-worker-capability-requirement/README.md
Roncossek: In addition to the current example of secure boot, lets imagine more features would arrive, Cloudprofiles would look something like this:

```yaml
  - name: gardenlinux
    versions:
      - version: 2150.4.0
        capabilityFlavors:
          - architecture: [amd64]
      - version: 2150.4.0-inplace
        inPlaceUpdates: { supported: true }
        fipsSupported: false
        capabilityFlavors:
          - architecture: [amd64]
      - version: 2150.4.0-fips
        fipsSupported: true
        capabilityFlavors:
          - architecture: [amd64]
      - version: 2150.4.0-fips-inplace
        inPlaceUpdates: { supported: true }
        fipsSupported: true
        capabilityFlavors:
          - architecture: [amd64]
```

In addition to adding a semver multiple times, new APIs would have to be added in several places. Also each field in the image version API would also need to implement its own filter for image selection and admission.

rfranzke: Please add this to the document :)

geps/0054-worker-capability-requirement/README.md
rfranzke: Yes, something like your second proposal is what I was having in mind. This comes with compatibility complications, though, and would indeed require something like your first proposal if introduced in the current API version. Not too nice either. 

Giving it a second thought now, probably the more pragmatic approach is what you already proposed: `gardener-`-prefixed reserved, well-known names. We have to take care to apply the same API conventions to those when it comes to adding/changing/removing them.

geps/0054-worker-capability-requirement/README.md
Roncossek: Added it to the Motivation section, it might be a bit long now, but I guess thats fine, if otherwise the Motivation is not motivating the proposal 😃

geps/0054-worker-capability-requirement/README.md
Roncossek: Then I added the part as another alternative, is that fine? 

I also noted, that this would be a section to pay attention to once a new API version is in the making.

Roncossek: Currently most of the features in the example are not yet fully qualified. So its not sure if they all will require a gardener or worker owned feature. So if the gardener owned capabilities feels unclean or unjustified. It might also be an option to reserve the only `gardener-*` namespace for now. This reservation gives us options for the future without committing to a final design yet.

geps/0054-worker-capability-requirement/README.md
```diff
@@ -0,0 +1,258 @@
...
+    machineCapabilities:
+      gardenerOwned:
+        nodeUpdateTypes: ["rolling", "in-place"]
+        bootTypes: ["secure", "legacy"]
+      custom:
+        - name: architecture
+          values: ["amd64", "arm64"]
+  ```
+
+  Internally these would still be encoded as capabilities so the matching algorithm is unaffected. Rejected because it is a breaking change to `spec.machineCapabilities`, every new Gardener-owned feature would still require a CloudProfile API change (defeating one of GEP-33's goals), and the prefix approach gives the same orchestration guarantees once the same API conventions are applied to reserved `gardener-*` names. A future API version can still promote individual reserved capabilities to first-class fields.
```
rfranzke: > "every new Gardener-owned feature would still require a CloudProfile API change"

I'd remove this - this is still the case with your proposal (introducing a new constant for the `gardener-*`) prefix is the same as adding a new field (from API compatibility PoV).

rfranzke (commented):
/cc @timebertt

geps/0054-worker-capability-requirement/README.md
Roncossek: Thats true! I updated the section.

geps/0054-worker-capability-requirement/README.md
ScheererJ: Yes, that should work.

Issue #63, opened 2026-05-12
Diki extension


AleksandarSavchev:
### Enhancement Description

- One-line enhancement description: Gardener extension for diki
- Steering: 
- Gardener Enhancement Proposal:  https://github.com/gardener/enhancements/pull/64
- Tracking Issue: 
- GEP (`g/enhancements`) update PR(s):

_Please keep this description up to date. This will help everybody to track the evolution of the enhancement efficiently._

PR #64, opened 2026-05-12
GEP-0063: Diki Extension

AleksandarSavchev:
- How to categorize this PR:
/area security
/area compliance
/kind enhancement

- One-line PR description: Add GEP for `Diki Extension` (gardener extension that would deploy the [diki-operator](https://github.com/gardener/diki-operator))

- Issue link: https://github.com/gardener/enhancements/issues/63

- Other comments:


geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+    - [API](#api)
+      - [ComplianceScan](#compliancescan)
+      - [ReportOutput](#reportoutput)
+      - [ScheduledComplianceScan](#scheduledcompliancescan)
+    - [Components](#components)
+      - [diki-extension-controller](#diki-extension-controller)
+      - [diki-operator](#diki-operator)
+      - [diki-admission-controller](#diki-admission-controller)
+      - [diki-run](#diki-run)
+    - [Architecture](#architecture)
```
vpnachev: Looks like there is no paragraph named `Architecture`

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+  users must collect, store, and distribute themselves.
+- No cluster-native integration: There is no Kubernetes-native way to
+  request a scan, inspect results, or configure recurring scans — everything
+  happens outside the cluster API.
+
+Providing compliance scanning as a first-class Gardener extension removes these
+burdens and makes compliance accessible to all shoot users through familiar
+Kubernetes resource semantics.
+
+### Goals
```
vpnachev, line 47: I think it will be of common interest to scan not only end-user shoot clusters, but also Seeds, Gardens and probably self-hosted shoots. Could you extend the proposal so that the extension covers also these use cases?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+
+-  Remediation of compliance findings. Diki is a detective tool, it reports
+   non-compliance but does not modify cluster resources to fix findings.
+-  Deploying or managing persistent storage backends for reports (e.g.,
+   PostgreSQL, OpenSearch, ODG). The `diki-operator` will support exporting
+   reports to such storage, but provisioning and operating the storage itself is
+   out of scope.
+-  Integration with the Gardener Dashboard.
+
+
```
vpnachev: You can remove one of the two empty lines here.
```suggestion
```

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+Providing compliance scanning as a first-class Gardener extension removes these
+burdens and makes compliance accessible to all shoot users through familiar
+Kubernetes resource semantics.
+
+### Goals
+
+-  Introduce `gardener-extension-diki` as a Gardener extension
+   that deploys the `diki-operator` into shoot control planes on seeds.
+-  Allow shoot users to run on-demand compliance scans by creating a
```
vpnachev: Do you envision to support configuring the scans and report outputs via the Gardener API? Admins in the shoot clusters can interfere with the diki CRDs and misconfigure them, Gardener users may want to define only once the rulesets for all of their shoots.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+1. Deploys the `diki-operator` into the shoot's namespace on the seed.
+2. Applies the three CRDs (`ComplianceScan`, `ReportOutput`,
+   `ScheduledComplianceScan`) and required RBAC resources to the shoot cluster
+   via `ManagedResource`.
+
+Users interact exclusively through the shoot API server — creating
+`ComplianceScan` or `ScheduledComplianceScan` resources. The `diki-operator` in
+the seed watches these resources (via the shoot API server) and orchestrates
+scan execution by creating `diki-run` Jobs in the shoot's control-plane namespace.
+The `diki-run` Pod writes summary reports back to the `ComplianceScan` status
```
vpnachev, line 91: Will diki scan only the data plane? Considering it is running in the control plane and have access to the control plane components, will it be an option for the end-users to get reports about their control plane compliance posture?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+![Diki service](diki-service.png)
+
+### Notes/Constraints/Caveats
+
+- `diki-operator` is under active development. The
+  [diki-operator repository](https://github.com/gardener/diki-operator) is in
+  early development. API shapes and component boundaries described in this GEP
+  may change as the implementation matures. The GEP captures the target design.
+
+- Scan execution happens on the seed, not in the shoot data plane. The
```
vpnachev, line 103: With what permissions and access in the seed is the execution job running? Should it be restricted in terms of network, permissions, privileges, etc?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+The extension is registered via `ControllerDeployment` and
+`ControllerRegistration`:
+
+```yaml
+apiVersion: core.gardener.cloud/v1beta1
+kind: ControllerDeployment
+metadata:
+  name: gardener-extension-diki
+helm:
+  rawChart: <base64-encoded Helm chart>
```
vpnachev: nit: You can use the `v1` version and refer to the chart as OCI image.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+The extension introduces three CRDs under the `diki.gardener.cloud` API group.
+All resources are cluster-scoped in the shoot cluster.
+
+#### ComplianceScan
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ComplianceScan
+metadata:
+  name: example-compliancescan
```
vpnachev: nit, it is not a common practice to put the kind of the resource in the name. 


```suggestion
  name: example
```

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+spec:
+  dikiVersion: v0.24 # defaults to latest available minor version
```
vpnachev, line 191: Is this the diki CLI version? How is this version verified by the diki-operator?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+      summary:
+        passed: 42
+        failed: 3
+        warning: 1
+        errored: 0
+        skipped: 2
+        accepted: 0
+      rules:
+        failed:
+        - id: "V-242381"
```
vpnachev, line 263: Considering the limit of ~1MB per resource in the ETCD, I am not sure it is a good idea to include details about the rules, the summary is fine, but detailed report might cause issues.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+          name: The Kubernetes API Server must have an audit policy set.
+  - id: security-hardened-k8s
+    ...
+```
+
+The `ComplianceScan` resource represents a single compliance scan execution.
+Its spec is immutable after creation. The `spec.rulesets` field allows users to
+select which rulesets to include and at which version; it defaults to all
+available rulesets at their latest versions. Ruleset and rule options are
+supplied via `ConfigMap` references. The `spec.outputs` field references
```
vpnachev, line 273: What strategy is applied when multiple outputs are configured, e.g. is it expected all of them to successfully accept the reports to consider the scan as successful, or is it enough just one/few of them to succeed?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+  output:
+    configMap:
+      namePrefix: compliance-scan-report-
+```
+
+The `ReportOutput` resource defines a storage destination for detailed
+compliance reports. It is immutable and can be referenced by multiple
+`ComplianceScan` resources. Each `ReportOutput` defines exactly one output
+type. The initial implementation supports `ConfigMap`-based output; additional
+outputs will be added in future iterations (e.g., PostgreSQL, OpenSearch).
```
vpnachev, line 302: Out of cluster outputs most probably will need credentials to authenticate with the output, should such credentials be configured via the Gardener API or they will be configured in the shoot?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+
+The `ScheduledComplianceScan` resource allows users to define recurring
+compliance scans following the `CronJob`/`Job` pattern. The operator creates
+`ComplianceScan` resources according to the cron schedule and manages their
+lifecycle. History limits control how many completed and failed `ComplianceScan`
+resources are retained.
+
+### Components
+
+![Sequence diagram](sequence-diag.png)
```
vpnachev, line 422: `Cluster API` could be confused with https://cluster-api.sigs.k8s.io/, probably the better term is `Kubernetes API`

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+- Watch `ComplianceScan` resources via the shoot API server.
+- Watch `ScheduledComplianceScan` resources and create `ComplianceScan`
+  resources on schedule.
+- Launch `diki-run` Jobs in the shoot's seed namespace to execute scans.
+- Watch `ComplianceScan` status for scan completion (status is written by the
+  `diki-run` Job).
+
+#### diki-admission-controller
+
+Runs as part of the `diki-operator`.
```
vpnachev: As it is running as part of the diki-operator, I think it does not need a dedicated paragraph, instead its responsibilities can be written in the diki-operator paragraph.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+`ComplianceScan` status in the shoot cluster (via the shoot API server) with
+summary results (per-ruleset pass/fail counts, failed rule lists, and output
+references).
+
+### Lifecycle Management
+
+| Phase | Behaviour |
+|-------|-----------|
+| Reconcile | Deploys `diki-operator` and diki-admission-controller to the shoot namespace on the seed. Creates a `ManagedResource` with CRDs, RBAC, and supporting resources for the shoot cluster. Waits for `ManagedResource` health before marking the `Extension` as reconciled. |
+| Delete | Deletes the `diki-operator` deployment and the `ManagedResource`. Waits for all managed objects to be removed from the shoot cluster before completing. |
```
vpnachev: Will it delete also the custom resource definitions for the `ComplianceScan`, `ScheduledComplianceScan`, `ReportOutput`, etc.?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+
+## Future Enhancements
+
+- Security-hardened shoot cluster ruleset: Diki includes a
+  [security-hardened shoot cluster](https://github.com/gardener/diki/tree/main/docs/rulesets/security-hardened-shoot-cluster)
+  ruleset that evaluates Gardener-specific security properties. Some rules in
+  this ruleset require access to the garden cluster (e.g., to inspect Shoot and
+  CloudProfile resources). Supporting this ruleset will require a mechanism for
+  the diki-run Job to obtain scoped, read-only garden cluster credentials —
+  a design that is deferred to a future iteration.
```
vpnachev, line 520: There is already a concept how extensions to get access to the Gardener API, however I am not sure diki really need it, both the shoot and cloudprofile specs are available in the `extensions.gardener.cloud/v1alpha1.Cluster` resource. The diki-extension (or the operator) can read the resource and extract the Shoot and the CloudProfile out of it.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+- Version management: Expose available Diki versions and ruleset versions
+  to users via the API.
+- Persistent storage backends: Add support for PostgreSQL, OpenSearch, and
```
vpnachev: Do you envision an option for non-solution specific output? Such solution should satisfy the need of those users that are using not widely adopted storage products.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+  ODG as report output destinations, removing the ConfigMap size limitation.
+- Dashboard integration: Integrate with the Gardener Dashboard to
+  visualize compliance scan summary results, and allow users to trigger
+  scans from the UI.
+- Report access proxy: Provide an authenticated proxy or server for
```
vpnachev, line 530: Is this scoped only to reports stored within the Seed or Gardener installation, e.g. managed storage?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,493 @@
...
+  users may find the additional API types unexpected if they do not actively use
+  compliance scanning.
+
+## Alternatives
+
+### Central Compliance Service in a Dedicated Cluster
+
+An alternative architecture would deploy the compliance CRDs
+(`ComplianceScan`, `ReportOutput`, `ScheduledComplianceScan`) as namespaced
+resources in the garden cluster, while running the compliance operator in a
```
vpnachev, line 549: I would still find it valuable if I can configure some scan jobs centrally and rely they are applied to all of my shoots or keep the configuration together with the shoot spec.

vpnachev (commented):
Thanks for the proposal, very well structured!

My impression is that it is covering quite some diki-operator specifics like the CRDs and how the operator is running the scans, but this proposal tries to focus on the extension. Should the diki-operator become a detailed topic of this GEP, or does it need a dedicated one, or keep its development without a GEP (and try to clean up this extension from operator specific topics)?

geps/0063-diki-extension/README.md
AleksandarSavchev: I agree, it would tremendously increase the value of the extension if it can cover all Gardener scenarios. However, the main motivation of this proposal is to help the end users with the execution of compliance scans.

I have added it to Future Enhancements

geps/0063-diki-extension/README.md
AleksandarSavchev: The initial plan was to add the diki CRDs to the garden cluster. This idea was dropped with the main concern being increased load to the Garden API. Even if the load increase isn't high from the diki CRDs, it will encourage other teams/ components to deploy their CRDs in the garden cluster.

Gardener users will need to create `ScheduledComplianceScan`s per Shoot. The ruleset options `ConfigMap` can be created in the project and referenced by `Shoot` clusters.

geps/0063-diki-extension/README.md
AleksandarSavchev: Currently the proposal is to scan only the data plane.

We can of course scan the control plane as well, however most of the rules we would check in the control plane are the responsibility of Gardener and not the cluster owner. For the rules, where the cluster owners would be responsible, we have a ruleset that checks the Shoot manifest (https://github.com/gardener/diki/blob/main/docs/rulesets/security-hardened-shoot-cluster/ruleset.md). The ruleset is not part of the proposal, since it requires access to the garden cluster to do it's checks. As you have mentioned bellow, one solution would be to use the Cluster resource on the seed.

geps/0063-diki-extension/README.md
AleksandarSavchev: To run the diki scan on the shoot the execution job would need the following ClusterRole: https://github.com/gardener/diki/blob/main/example/rbac/managedk8s.yaml.

Additionally the diki-exporter would need more permissions/ credentials depending on the configuration (create Configmap, write access to Postgres etc.)

I have added this note.

geps/0063-diki-extension/README.md
AleksandarSavchev: The operator will have a list of supported diki versions. We will only support 1 version per minor version and would keep the supported version to a limited amount.

The plans on how to expose the supported versions to end users are added as future enahncement.

geps/0063-diki-extension/README.md
AleksandarSavchev: Agreed, the idea here is only to add "findings" (failed/ errored/ warning) with only their names and IDs, without any other details.

geps/0063-diki-extension/README.md
AleksandarSavchev: If one output fails the ComplianceScan should get into Failed phase. It would be hard for the end user to notice that there was an error with one ore more of the outputs if we mark it as succeeded.

Such failures would most likely be caused by misconfiguration of the outputs or downtime of the persistent storage. We plan to add a retry mechanism for failed outputs, which can lower the changes if this happening.

I am open for suggestions on this case.

geps/0063-diki-extension/README.md
AleksandarSavchev: They would live as secrets referenced by ReportOutput CRs in the shoot.

geps/0063-diki-extension/README.md
AleksandarSavchev: I think we should not delete these CRDs since the user is the one who has created them and should be responsible for them. Also, in the instance where the user disables the extension and then re-enables it, he should expect that his previously configured resources would still be present.

I am also up for discussion here.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, there should be the option to push the diki report to a custom endpoint. I have added it in the future enhancements.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, I do not think we should be actively reading from end user storage.

AleksandarSavchev:
> My impression is that it is covering quite some diki-operator specifics like the CRDs and how the operator is running the scans, but this proposal tries to focus on the extension. Should the diki-operator become a detailed topic of this GEP, or does it need a dedicated one, or keep its development without a GEP (and try to clean up this extension from operator specific topics)?

Good point, it does go into operator details. I'd still prefer to keep them here though, since the diki-operator does not have its own dedicated GEP and is being developed as part of this effort. The CRDs and scan orchestration are essentially the end user story. They define how shoot users actually interact with compliance scanning. I feel like here is the appropriate place to get feedback and discuss the compliance API.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+## Summary
+
+[Diki](https://github.com/gardener/diki) is a compliance checker that evaluates
+the security posture of Kubernetes clusters against pluggable rulesets such as
+the [DISA Kubernetes STIG](https://public.cyber.mil/stigs/). Today, Gardener
+users must install and operate the Diki CLI themselves — scheduling scans,
+managing configuration, and collecting reports manually.
+
+This GEP proposes a new Gardener extension, `gardener-extension-diki`, that deploys
```
ScheererJ, line 11: Should this be `gardener-extension-shoot-diki`? Gardener extensions are somewhat inconsistent in this regard, though.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+  top of defaults by providing their own ruleset and rule options via
+  `ConfigMap` references in the `ComplianceScan` spec.
+
+### Risks and Mitigations
+
+| Risk | Likelihood | Impact | Mitigation |
+|------|-----------|--------|------------|
+| `diki-operator` API changes during development | High | Medium | The extension will track the operator's API as it stabilizes. CRD versioning (`v1alpha1`) signals instability to users. |
+| Scan Jobs consume excessive seed resources | Medium | Low | The `diki-operator` creates one `diki-run` Job per scan. `ScheduledComplianceScan` history limits (`successfulScansHistoryLimit`, `failedScansHistoryLimit`) bound the number of retained scan resources. Concurrent Jobs can be limited. |
+| CRDs in shoot clusters add API surface users may not expect | Low | Low | CRDs are only installed when the extension is explicitly enabled on the shoot. |
```
ScheererJ, line 130: Experience with kyverno has shown that users can easily overload their `etcd` with kyverno reports. Do we face the same issue here or do we ensure somehow that the volume is somehow limited?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+## Design Details
+
+### Extension Registration
+
+The extension is registered via `ControllerDeployment` and
+`ControllerRegistration`:
```
ScheererJ: It should probably also support the `Extension` API of `operator.gardener.cloud/v1alpha1`.

geps/0063-diki-extension/README.md
ScheererJ: Why should the user specify a version at all?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ComplianceScan
+metadata:
+  name: example
+spec:
+  dikiVersion: v0.24 # defaults to latest available minor version
+  rulesets:
+    - id: disa-kubernetes-stig
+      version: v2r4
```
ScheererJ, line 194: Where is the ruleset version coming from? How does a user know what to specify here?

geps/0063-diki-extension/README.md
ScheererJ: What is expected/worst case amount of issues here? Can users blow up their `etcd` with this?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+      rules:
+        failed:
+        - id: "V-242381"
+          name: The Kubernetes API Server must have an audit policy set.
+  - id: security-hardened-k8s
+    ...
+```
+
+The `ComplianceScan` resource represents a single compliance scan execution.
+Its spec is immutable after creation. The `spec.rulesets` field allows users to
```
ScheererJ, line 270: Are the referenced `ConfigMap`s also supposed to be immutable? Otherwise the meaning of the spec may change over time when the referenced `ConfigMap`s are changed.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ReportOutput
+metadata:
+  name: example-configmap-output
+spec:
+  output:
+    configMap:
+      namePrefix: compliance-scan-report-
```
ScheererJ, line 295: What else can be specified at `spec` level? `output` seems somewhat redundant as I would have expected that `ReportOutput` is about specifying outputs. In other words, should we remove the `output` level to have a simpler CRD?
WDYT?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ScheduledComplianceScan
+metadata:
+  name: weekly-compliance-scan
+spec:
+  schedule: "0 0 * * 0" # cron expression, defaults to weekly on Sunday at midnight
+  successfulScansHistoryLimit: 3
+  failedScansHistoryLimit: 1
```
ScheererJ: This example seems counter-intuitive. Would you not want to keep more failed results than successful ones?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+    name: weekly-compliance-scan-28459230
+  lastScheduleTime: "2025-12-28T00:00:00Z"
+  lastCompletionTime: "2025-12-28T00:12:34Z"
+```
+
+The `ScheduledComplianceScan` resource allows users to define recurring
+compliance scans following the `CronJob`/`Job` pattern. The operator creates
+`ComplianceScan` resources according to the cron schedule and manages their
+lifecycle. History limits control how many completed and failed `ComplianceScan`
+resources are retained.
```
ScheererJ: What does retention mean? Will it delete the `ComplianceScan` resources only or also the output, e.g. `ConfigMap`s (if possible)?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+### Components
+
+![Sequence diagram](sequence-diag.png)
+
+#### diki-extension-controller
+
+Runs on the seed as a standard Gardener extension controller. Responsibilities:
+
+- Watch `Extension` objects of type `diki`.
+- Deploy the `diki-operator` to the shoot's seed namespace.
```
ScheererJ: Will the `diki-operator` not also be deployed via `ManagedResource`? If not, why not?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+  `ManagedResource`.
+
+#### diki-operator
+
+Runs in the shoot's seed namespace. Responsibilities:
+
+- Watch `ComplianceScan` resources via the shoot API server.
+- Watch `ScheduledComplianceScan` resources and create `ComplianceScan`
+  resources on schedule.
+- Launch `diki-run` Jobs in the shoot's seed namespace to execute scans.
```
ScheererJ, line 441: Who checks for the `diki-run` jobs at runtime? It sounds like `diki-operator` just triggers the job in a fire-and-forget model and does not care whether it fails or is successful. What happens if the job crashes?
Is job meant here as `Job`, i.e. kubernetes resource, or is it just meant as a `Pod` to execute something?

geps/0063-diki-extension/README.md
ScheererJ: +1 for `Cluster` resource usage instead of Gardener API credentials.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+- Security-hardened shoot cluster ruleset: Diki includes a
+  [security-hardened shoot cluster](https://github.com/gardener/diki/tree/main/docs/rulesets/security-hardened-shoot-cluster)
+  ruleset that evaluates Gardener-specific security properties. Some rules in
+  this ruleset require access to the garden cluster (e.g., to inspect Shoot and
+  CloudProfile resources). Supporting this ruleset will require a mechanism for
+  the diki-run Job to obtain scoped, read-only garden cluster credentials —
+  a design that is deferred to a future iteration.
+- Version management: Expose available Diki versions and ruleset versions
+  to users via the API.
```
ScheererJ, line 522: I understand that this is necessary if users need to specify the version, but I do not understand why users should do that. Could you please explain?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+- Persistent storage backends: Add support for PostgreSQL, OpenSearch, and
+  ODG as report output destinations, removing the ConfigMap size limitation.
+  Additionally, there should be an option to export to a custom HTTP endpoint.
```
ScheererJ: Is the `ConfigMap` size limitation a real limitation in reality?
Could you please link to the additional output targets, especially ODG as not everybody may be aware of it?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+- Report access proxy: Provide an authenticated proxy or server for
+  accessing stored compliance reports.
+- Extend the scope of the Diki extension to support the different Gardener
+  scenarios (e.g. seed cluster, garden cluster, etc.) and not only shoot clusters.
+
+## Drawbacks
+
+- Dependency on an in-development operator. The `diki-operator` is under
+  active development and its APIs are not yet stable. This can cause breaking
+  changes across versions.
```
ScheererJ: Are you referring to problems that could happen during extension development or are you implying that users may be impacted?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+- CRD footprint in shoot clusters. Enabling the extension adds three CRDs
+  to the shoot cluster API surface. While these are only installed on opt-in,
+  users may find the additional API types unexpected if they do not actively use
+  compliance scanning.
```
ScheererJ: I am not sure I understand this point? It seems like the extension needs to be actively enabled. Hence, it should not come as a surprise if enabling it causes CRDs to be deployed.
Please explain.

ScheererJ (commented):
Thanks for working on making cluster compliance easier than ever.

geps/0063-diki-extension/README.md
AleksandarSavchev: There can be breaking changes to the `diki` config, which users should address before upgrading to the new `diki` version.

In `diki` we try to keep breaking changes to the config to the minimum. However, such are still possible and we need to give users time to fix their configs.

geps/0063-diki-extension/README.md
AleksandarSavchev: Currently in the worst case there are ~50 rules that can fail, which would be ~100 additional lines. This should not blow up the `etcd`.

geps/0063-diki-extension/README.md
AleksandarSavchev: We went with `gardener-extension-diki` as the simplest name format for a Gardener extension.

Additionally, as suggested by @vpnachev [here](https://github.com/gardener/enhancements/pull/64#discussion_r3304824985), we would look into enabling the extension for the Garden cluster/ Seeds/ Soils where the `-shoot-` part would not make much sense.

geps/0063-diki-extension/README.md
AleksandarSavchev: Diki reports are lighter than kyverno reports, but the size depends on cluster size and number of findings.

For an empty Shoot cluster, a Diki report is around ~50 KB. Our largest reports reach ~10 MB raw, which compressed (gzip + base64) is ~500 KB.

That said, ConfigMap-based storage is intended as a simple solution for medium sized clusters and as an entry point for getting started with the diki-operator. For larger clusters, users should configure external outputs (e.g., PostgreSQL, OpenSearch).

I have added it as a risk.

geps/0063-diki-extension/README.md
AleksandarSavchev: The different diki versions may support different ruleset versions.

For example `diki` `v0.24.0` supports `disa` version `v2r3` and `v2r4`, while `diki` `v0.25.0` supports `disa` `v2r4` and `v2r5`. The rule IDs are stable between versions, but there might be updated, new and / or removed rules.

If a user configures a non-supported version of a ruleset, there will be a validation error with the supported versions.

There have also been talks about exposing the diki and ruleset versions to users, it is added to future enhancements.

geps/0063-diki-extension/README.md
AleksandarSavchev: The referenced `ConfigMap`s are not expected to be immutable.

> Otherwise the meaning of the spec may change over time when the referenced ConfigMaps are changed.

The scan uses the rule options are in the ConfigMap at execution time. In the Diki report, we don't persist the exact options used, but we do record which findings were accepted/silenced by them. The user can see the effect of the rule options without needing the exact configuration.

geps/0063-diki-extension/README.md
AleksandarSavchev: I agree, currently the `output` field is redundant. I would prefer to leave it there, since in the future we might decide to add additional configurations (e.g report format `json`/`html`).

geps/0063-diki-extension/README.md
AleksandarSavchev: Currently the outputs are not deleted with the `ComplianceScan`. Compliance reports should be kept for audit traceability. However, retaining `ConfigMap`s might become a problem for the etcd if the user does not clean them up.

We might think about adding a configuration to delete the report `ConfigMap` with the `ComplianceScan` or another solution to this problem

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, fixed the wording.

geps/0063-diki-extension/README.md
AleksandarSavchev: The `diki-operator` is responsible for updating the phase and conditions of the `ComplianceScan`, while the `report-exporter` only fills in the `status.outputs` field.

On a failed `Job` execution the operator will mark the `ComplianceScan` as `failed` with appropriate conditions.

The operator does not directly create a `Pod` but instead leverages a kubernetes `Job`. All other required resources (e.g. `ConfigMap`s with diki configurations) created for the compliance scan have the diki-scan `Job` in `ownerRefference`.

The document has been updated to include this responsibility.

geps/0063-diki-extension/README.md
AleksandarSavchev: As explained in an above comment different `diki` version support different ruleset versions.

Additionally there might be breaking changes to the `diki` config, which users should address before upgrading to the new `diki` version.

geps/0063-diki-extension/README.md
AleksandarSavchev: In most cases the size limitation should not be a problem, but I think it should be noted.

I have added the requested links to the documentation.

geps/0063-diki-extension/README.md
AleksandarSavchev: This concern is regarding extension development.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, it is an expected behaviour.

The question is if we consider it a drawback the addition of 3 new CRDs to the cluster? If not, i can remove this one, since I do not have a string opinion on it.

geps/0063-diki-extension/README.md
ScheererJ: How can users figure out what versions are supported/available? Is there an API?

geps/0063-diki-extension/README.md
ScheererJ: How should a user in reality get the available IDs and versions? It cannot just be trial and error. Trying to create something with the sole intention to get the list of versions from the error message does not seem great.

geps/0063-diki-extension/README.md
ScheererJ: From my point of view the `ConfigMap` is an extension of the spec as it influences what is actually checked. Effectively, the configuration is not immutable if the `ConfigMap` is mutable. Is this the desired state, though? In other words, could you please explain why the spec is immutable and the `ConfigMap` not?

geps/0063-diki-extension/README.md
ScheererJ: Would the format not also relate to the `output`, i.e. would it really not be under `output`?

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, we cannot leave the trial and error solution as final.

The real solution should include an API, where users can get the supported version. We can have a new "singleton" CRD e.g. `DikiInfo` `DikiProfiles` where the versions will be listed.

I initially scoped this as a future enhancement, but if you deem it necessary for the initial implementation I can also add a proposal for it in this GEP.

geps/0063-diki-extension/README.md
AleksandarSavchev: The `ComplianceScan` spec is immutable because it represents a single scan execution. Once created, mutating the spec would have no effect on the results since the scan either already ran or is in progress.

The referenced ConfigMap, on the other hand, is intentionally mutable. A user may maintain a single options ConfigMap that they reference across multiple ComplianceScan runs throughout the cluster lifecycle. For example, as workloads change on the cluster, users would continuously update their rule options (e.g., adding accepted/silence rules for known findings) without needing to create a new ConfigMap for every scan.

The operator resolves the ConfigMap content at the time the scan starts. Subsequent changes to the `ConfigMap` do not affect an already running or completed scan, they only take effect for future scans that reference it.

I can see how this would be in conflict with the design of the `ComplianceScan`, however I would prefer the `ConfigMap`s to be left mutable considering how often it can be required by users to change silencing rules.

geps/0063-diki-extension/README.md
AleksandarSavchev: The output field is designed as a discriminated union, exactly one output type (e.g., `configMap`, `odg` ...) can be configured. The format is not specific to a particular output type, so it would belong as a top-level field in the spec alongside output, not nested under it. For example:

``` yaml
  spec:
    format: json
    output:
      configMap: ...
```

This keeps output focused purely on the destination configuration.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+    - [Components](#components)
+      - [diki-extension-controller](#diki-extension-controller)
+      - [diki-operator](#diki-operator)
+      - [diki-run](#diki-run)
+    - [Lifecycle Management](#lifecycle-management)
+  - [Future Enhancements](#future-enhancements)
+  - [Drawbacks](#drawbacks)
+  - [Alternatives](#alternatives)
+    - [Central Compliance Service in a Dedicated Cluster](#central-compliance-service-in-a-dedicated-cluster)
+    - [Trivy Operator for Compliance Scanning](#trivy-operator-for-compliance-scanning)
```
rfranzke: This is no longer part of the template.

geps/0063-diki-extension/README.md
rfranzke: +1 for `gardener-extension-diki`. I don't know why we started to put `shoot-` in there. 🫠

geps/0063-diki-extension/README.md
rfranzke: If you already agree, why can't we consider this right from the start? We don't have to ship it in v0.1, but it seems to be very reasonable to also offer scans for the garden and seed clusters.

geps/0063-diki-extension/README.md
rfranzke: Agree @AleksandarSavchev. I think it would be great to write this reasoning down.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+#### ComplianceScan
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ComplianceScan
+metadata:
+  name: example
+spec:
+  dikiVersion: v0.24 # defaults to latest available minor version
+  rulesets:
```
rfranzke, line 192: camelCase?

geps/0063-diki-extension/README.md
rfranzke: This is similar to what has been introduced with Falco, and other addons/extensions mighit require this as well. Perhaps we should take a step back and spend time with thinking this through generally (how can an extension provide configuration/versions for users (similar to `CloudProfile`s), and how can we handle auto-update/maintenance/force-upgrade/expirations, etc.)?

If we now introduce another shortcut with `DikiProfile` or what not, we open the door even further for the future, and migration effort back to a more generic solution will become harder in sum.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,491 @@
...
+    - id: disa-kubernetes-stig
+      version: v2r4
+      options:
+        ruleset:
```
rfranzke, line 196: camelCase?

geps/0063-diki-extension/README.md
AleksandarSavchev: I have now added the garden and seed cluster scans to the goals and have documented them in #Cluster Compatibility.

geps/0063-diki-extension/README.md
AleksandarSavchev: In diki we recognise `rulesets` as a single word. In different places I have seen it written as 2 words and as a single one.

geps/0063-diki-extension/README.md
AleksandarSavchev: Agreed. We would refrain from introducing a new `CloudProfile` for diki.

A generic mechanism for how extensions provide versions to users I think would be out of scope for this GEP, it can require its own dedicated GEP. When such a generic solution becomes available, we will adopt it.

geps/0063-diki-extension/README.md
AleksandarSavchev: I have documented this reason in the `#Alternatives` -> `#Central Compliance Service in a Dedicated Cluster` section of the GEP.

geps/0063-diki-extension/README.md
ScheererJ: Is the summary then that that the ruleset version for now appears out of thin air, i.e. there is no API for it, but at a later point in time when we have a generic ExtensionProfile (or however we will call it) this will be used?

geps/0063-diki-extension/README.md
ScheererJ: Sure, the `ConfigMap`s can stay mutable. Do we need to copy the content to the status/result, though? Then we would know with what kind of settings the `ComplianceScan` actually ran. Or is this just me overthinking it?

geps/0063-diki-extension/README.md
ScheererJ: Nit: Not sure if it then would make sense to also rename `output` to `destination`. However, I understand your reasoning now better.

geps/0063-diki-extension/README.md
ScheererJ: From my point of view, adding three CRDs for something a user actively configured is not a bad thing.

AleksandarSavchev: Ok, removed then in the latest commit https://github.com/gardener/enhancements/pull/64/changes/3c549d6be3ba7c82460a9bb487d74f749bafc457

geps/0063-diki-extension/README.md
AleksandarSavchev: I would prefer to stick to `output` as the CRD is named `ReportOutput`. The field described the `output` type.

geps/0063-diki-extension/README.md
AleksandarSavchev: I do not think we need to provide the exact ruleset and rule settings used, since skipped rules and exceptions are documented in the diki report.

geps/0063-diki-extension/README.md
AleksandarSavchev: Currently, yes.

geps/0063-diki-extension/README.md
vpnachev: One option that come to my mind is to classify the outputs as `Required` and `Optional`, a scan to be successful all required outputs must accept the report, any failure on an optional output should not result in a failed scan. Via configuration validation it should be ensured at least one of the outputs is of class `Required`.

geps/0063-diki-extension/README.md
vpnachev: Just for announcing supported versions and rulesets, an interim solution could be a configmap in the `gardener-system-public` namespace readable by all authenticated users similar to https://github.com/gardener/gardener/blob/v1.145.0/docs/operations/configmap.md

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+
+This GEP proposes a new Gardener extension, `gardener-extension-diki`, that deploys
+the [diki-operator](https://github.com/gardener/diki-operator) into shoot
+cluster control planes. The operator introduces three Custom Resource
+Definitions — `ComplianceScan`, `ReportOutput`, and `ScheduledComplianceScan` —
+that allow shoot users to run on-demand and scheduled compliance scans
+declaratively, with summary results reported directly in resource status fields.
+The extension follows the standard Gardener extension contract
+(`ControllerRegistration`, `ManagedResource`-based deployment, admission
+webhooks) and keeps the compliance workload inside the seed, avoiding
```
vpnachev: `compliance workload` - I am not sure we have such term, probably `extension workload` is a better wording?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+particularly those operating in regulated environments that require periodic
+proof of adherence to security baselines such as the DISA Kubernetes STIG.
+
+Currently, Gardener recommends the Diki CLI for running compliance scans. This
+places the full operational burden on each user:
+
+- Manual scheduling: Users must set up their own cron jobs or CI pipelines
+  to trigger scans on a recurring basis.
+- Configuration management: Each user independently manages Diki
+  configuration files, ruleset versions, and rule options.
```
vpnachev, line 36: You can add another point that they also have to keep the diki cli version up-to-date themselves.

geps/0063-diki-extension/README.md
vpnachev: If I am not mistaken, the shared cluster role is about the access in the shoot cluster, while I was wondering whether the diki-run job would need any access to the seed via its API server, e.g. reading the cluster resource? 

I guess the diki-run job is not responsible to write to the configured output, so maybe it can benefit from network policies allowing network access only to the shoot API server (and eventually the seed API server)?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+  top of defaults by providing their own ruleset and rule options via
+  `ConfigMap` references in the `ComplianceScan` spec.
+
+### Risks and Mitigations
+
+| Risk | Likelihood | Impact | Mitigation |
+|------|-----------|--------|------------|
+| `diki-operator` API changes during development | High | Medium | The extension will track the operator's API as it stabilizes. CRD versioning (`v1alpha1`) signals instability to users. |
+| Scan Jobs consume excessive seed resources | Medium | Low | The `diki-operator` creates one `diki-run` Job per scan. `ScheduledComplianceScan` history limits (`successfulScansHistoryLimit`, `failedScansHistoryLimit`) bound the number of retained scan resources. Concurrent Jobs can be limited. |
+| ConfigMap-based report output may strain etcd for large clusters | Medium | Medium | ConfigMap output is intended as a simple solution for small clusters. Reports are compressed (gzip + base64) to stay within the 1 MB etcd limit. For larger clusters, external outputs (e.g., PostgreSQL, OpenSearch) should be used. |
```
vpnachev: base64 encoding is known to increase the payload size by 33%, if configMap can deal with raw bytes (eventually via [BinaryData](https://github.com/kubernetes/kubernetes/blob/4f711496824163f8d6b2ceecdb58b418538fbf69/staging/src/k8s.io/api/core/v1/types.go#L8097-L8107)) you might consider to not encode the payload.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+        - garden
+      lifecycle:
+        reconcile: AfterKubeAPIServer
+        delete: BeforeKubeAPIServer
+  deployment:
+    deploymentRefs:
+      - name: gardener-extension-diki
+```
+
+The extension controller is deployed per seed and watches `Extension` objects of
```
vpnachev:
```suggestion
The extension controller is deployed per seed and watches `extensions.gardener.cloud/v1alpha1.Extension` objects of
```


just to avoid confusion with the operator Extensions.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+
+The extension declares `clusterCompatibility` for shoot, seed, and garden
+clusters. The behaviour per cluster type is as follows:
+
+- Shoot: The extension deploys the `diki-operator` into the shoot's namespace
+  on the seed. CRDs and RBAC are applied to the shoot cluster. Scans evaluate
+  only the data plane. This is the initial implementation target.
+- Seed: Works analogously to shoot. The extension deploys the `diki-operator`
+  and applies CRDs and RBAC to the seed cluster. Scans evaluate both the
+  control plane and the data plane of the seed.
```
vpnachev: I think this will not be implemented exactly like it is described because diki might not have access to evaluate the control plane. ManagedSeed is what diki can fully scan - I mean data and control plane. 

Maybe it will be (this is just a thought, in reality could be implemented differently) something like:
- diki-operator deployed in the `garden` namespace of the seed
- the `ComplianceScan` will refer to a shoot hosted in the seed
- the diki-scan job will scan both the control and data plane 


It would be nice if you can add a bit more clarity where is the operator deployed and how the scan job running.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+- Garden: Both the runtime cluster and the virtual garden cluster are
+  examined. The CRDs (`ComplianceScan`, `ReportOutput`, `ScheduledComplianceScan`)
+  are located in the runtime cluster. The `diki-operator` runs scans against
+  both the runtime and virtual garden API servers, evaluating both the
```
vpnachev: Does single `ComplianceScan` scan the both the runtime and the virtual cluster or dedicated `ComplianceScan`s will be required.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+    lastTransitionTime: "2025-12-31T23:59:59Z"
+    reason: ComplianceScanCompleted
+    message: "ComplianceScan completed successfully."
+  outputs:
+  - outputName: example-configmap-output
+    phase: Completed
+    details:
+      configMapRef:
+        name: compliance-scan-report-njdjv
+        namespace: kube-system
```
vpnachev, line 229: How will this look like in case of a non-configmap output, e.g. OpenSearch?

geps/0063-diki-extension/README.md
vpnachev: It would be nice to add sample for non-configmap output with credentials and eventually additional configuration options. Considering these outputs are not yet implemented, changing the API is completely acceptable, but it would be nice to have some example already today in the GEP.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,528 @@
...
+
+#### ScheduledComplianceScan
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ScheduledComplianceScan
+metadata:
+  name: weekly-compliance-scan
+spec:
+  schedule: "0 0 * * 0" # cron expression, defaults to weekly on Sunday at midnight
```
vpnachev, line 393: To avoid pick load at the same time, it will be nice the default to have some random jitter, e.g. 15 minutes.

geps/0063-diki-extension/README.md
vpnachev: I think deleting the configmap should be the default behavior, especially because the configmap output is intended for demo/non-productive usage.

geps/0063-diki-extension/README.md
vpnachev: Considering the CRDs are deployed via managed resource, when the MR is deleted, the CRDs, unless explicitly configured, will be also deleted, which will trigger the deletion of the CRs. 

Orphaning the diki  CRDs could lead to users running the diki-operator on their own, however they can wrongly expect the extension to update the CRD definitions even when the extension is disabled. I would vote all diki related resources to be deleted when the extension is disabled. This can be changed later if users start asking the CRDs to be kept in the cluster.

geps/0063-diki-extension/README.md
AleksandarSavchev: Updated docu with https://github.com/gardener/enhancements/pull/64/commits/497d53be6e135540dc4c7bab74605488c31cd860

geps/0063-diki-extension/README.md
AleksandarSavchev: Added `optional` (defaults to `false`) field with https://github.com/gardener/enhancements/pull/64/commits/497d53be6e135540dc4c7bab74605488c31cd86

geps/0063-diki-extension/README.md
AleksandarSavchev: I have added example ReportOutputs for `PostgreSQL`, `OpenSearch` and `Webhook` (custom endpoint). As mentioned, their apis are volatile, will most likely be changed in the future.

geps/0063-diki-extension/README.md
AleksandarSavchev: I agree that for configMap it does make sense to not persist the reports. I have now added a note to the ReportOutput section about cleaning up configMap reports with the deletion of the complianceScan https://github.com/gardener/enhancements/pull/64/commits/497d53be6e135540dc4c7bab74605488c31cd860

geps/0063-diki-extension/README.md
AleksandarSavchev: Added a separate point for Seed (soils) which should cover this concern with https://github.com/gardener/enhancements/pull/64/commits/497d53be6e135540dc4c7bab74605488c31cd860

geps/0063-diki-extension/README.md
AleksandarSavchev: A single `ComplianceScan` can be configured for both the runtime and virtual cluster (there would need to be additional configuration on the `diki-config` on how to handle different rulesets).

The human operator would need to decide if he want to have a single `ComplianceScan` resource or multiple.

geps/0063-diki-extension/README.md
AleksandarSavchev: Added example for the new `ReportOutput`s added with https://github.com/gardener/enhancements/commit/497d53be6e135540dc4c7bab74605488c31cd860

geps/0063-diki-extension/README.md
AleksandarSavchev: Thank you for sharing this idea.

There is also the possibility for the `diki-operator` to deploy this configmap in the target cluster. Then we would have a solution outside the extension for users that would deploy the `diki-operator` manually (without extension/ outside of gardener).

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,638 @@
...
+  name: example-postgresql-output
+spec:
+  output:
+    postgresql:
+      url: "postgres.garden.svc:5432"
+      database: "diki_reports"
+      table: "compliance_reports"
+      secretRef:
+        name: postgresql-credentials
+        namespace: kube-system
```
vpnachev: Let's make the credentials configuration more robust. 

```suggestion
      credentialsRef:
        apiVersion: v1
        kind: Secret
        name: postgresql-credentials
        namespace: kube-system
```

CredentialsRef this way is open for new implementations without changing the API. One option that come to my mind is `v1.ServiceAccount` which token can be used for authentication, eventually an opaque config option will be required (to control the `audience` claim value and the duration of the token). For the first version implementing just support for `v1.Secret` should be enough.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,638 @@
...
+      tls:
+        enabled: true
+        caSecretRef:
+          name: postgresql-ca
+          namespace: kube-system
```
vpnachev, line 332:
```suggestion
        caSecretRef:
          name: postgresql-ca
          namespace: kube-system
          key: ca.crt
```

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,638 @@
...
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ReportOutput
+metadata:
+  name: example-webhook-output
+spec:
+  output:
+    webhook:
+      url: "https://compliance-api.corp.example.com/v1/reports"
+      headerSecretRef:
```
vpnachev:
```suggestion
      credentialsRef:
```

let stay consistent (use `secretRef` if the proposal for `credentialsRef` is not acceptable).

vpnachev (commented):
Thanks!
I have just few suggestions for the latest API proposal.

geps/0063-diki-extension/README.md
AleksandarSavchev: Agreed, added suggestion with https://github.com/gardener/enhancements/pull/64/changes/22697156e7288e771649792b34e03e3bba47b9d2

vpnachev (commented):
/lgtm

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+  top of defaults by providing their own ruleset and rule options via
+  `ConfigMap` references in the `ComplianceScan` spec.
+
+### Risks and Mitigations
+
+| Risk | Likelihood | Impact | Mitigation |
+|------|-----------|--------|------------|
+| `diki-operator` API changes during development | High | Medium | The extension will track the operator's API as it stabilizes. CRD versioning (`v1alpha1`) signals instability to users. |
+| Scan Jobs consume excessive seed resources | Medium | Low | The `diki-operator` creates one `diki-run` Job per scan. `ScheduledComplianceScan` history limits (`successfulScansHistoryLimit`, `failedScansHistoryLimit`) bound the number of retained scan resources. Concurrent Jobs can be limited. |
+| ConfigMap-based report output may strain etcd for large clusters | Medium | Medium | ConfigMap output is intended as a simple solution for small clusters. Reports are compressed (gzip) and stored via `BinaryData` to stay within the 1 MB etcd limit. For larger clusters, external outputs (e.g., PostgreSQL, OpenSearch) should be used. |
```
timebertt, line 129: The ConfigMap report result is created in the shoot, correct?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+| CRDs in shoot clusters add API surface users may not expect | Low | Low | CRDs are only installed when the extension is explicitly enabled on the shoot. |
+
+## Design Details
+
+### Extension Registration
```
timebertt, line 134: The example YAMLs in this section bloat the proposal. You could condense it to a short bullet point list:
* the extension supports all three cluster types: `garden`, `seed`, and `shoot`
* the envisioned lifecycle policy is `AfterKubeAPIServer` for reconcile and migrate operations and `BeforeKubeAPIServer` for delete operations

Apart from that, there is no new information in this section

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+  resources:
+    - kind: Extension
+      type: diki
+      globallyEnabled: false
+      clusterCompatibility:
+        - shoot
+        - seed
+        - garden
+      lifecycle:
+        reconcile: AfterKubeAPIServer
```
timebertt: Is `migrate` left out on purpose here?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+The extension declares `clusterCompatibility` for shoot, seed, and garden
+clusters. The behaviour per cluster type is as follows:
+
+- Shoot: The extension deploys the `diki-operator` into the shoot's namespace
+  on the seed. CRDs and RBAC are applied to the shoot cluster. Scans evaluate
+  only the data plane. This is the initial implementation target.
+- Seed (ManagedSeed): Works analogously to shoot. The extension deploys the
+  `diki-operator` in the control plane namespace and applies CRDs and RBAC to the
+  cluster. Scans evaluate both the control plane and the data plane of the seed.
+- Seed (Soil): The `diki-operator` is installed in the `garden` namespace of
```
timebertt: "soil" is not a well-defined term in Gardener OSS. You might want to refer to Seeds that are not backed by Shoots/ManagedSeeds instead.
Apart from that, I suggest just referring to `Seed` as this is the object where the extension is enabled.
You can still add a note that operators might want to prefer enabling the extension on the shoot for seeds backed by ManagedSeeds because that will include a control plane scan in contrast to enabling the extension on the Seed.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+
+The extension declares `clusterCompatibility` for shoot, seed, and garden
+clusters. The behaviour per cluster type is as follows:
+
+- Shoot: The extension deploys the `diki-operator` into the shoot's namespace
+  on the seed. CRDs and RBAC are applied to the shoot cluster. Scans evaluate
+  only the data plane. This is the initial implementation target.
+- Seed (ManagedSeed): Works analogously to shoot. The extension deploys the
+  `diki-operator` in the control plane namespace and applies CRDs and RBAC to the
+  cluster. Scans evaluate both the control plane and the data plane of the seed.
```
timebertt: I suggest removing this point. I understand that there is no difference from regular shoots. I.e., the operator enables the `diki` extension in the shoot backing the seed and there is no information about it in either `Seed` or `ManagedSeed` object.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+- Seed (Soil): The `diki-operator` is installed in the `garden` namespace of
+  the soil. CRDs and RBAC are applied to the soil cluster. Scans evaluate
+  the data plane.
+- Garden: Both the runtime cluster and the virtual garden cluster are
```
timebertt, line 163: What happens if the garden runtime cluster is also used as a seed cluster?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+metadata:
+  name: example-configmap-output
+spec:
+  output:
+    configMap:
+      namePrefix: compliance-scan-report-
+```
+
+The `ReportOutput` resource defines a storage destination for detailed
+compliance reports. It is immutable and can be referenced by multiple
```
timebertt, line 299: Why are `ReportOutput` objects immutable?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+
+#### ScheduledComplianceScan
+
+```yaml
+apiVersion: diki.gardener.cloud/v1alpha1
+kind: ScheduledComplianceScan
+metadata:
+  name: weekly-compliance-scan
+spec:
+  schedule: "0 0 * * 0" # cron expression, defaults to weekly on Sunday at midnight
```
timebertt, line 393: You might want to support specifying a time zone for the schedule instead of relying on the diki-operator's time zone

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+
+Runs on the seed as a standard Gardener extension controller. Responsibilities:
+
+- Watch `Extension` objects of type `diki`.
+- Deploy the `diki-operator` to the shoot's seed namespace via
+  `ManagedResource`.
+- Apply CRDs, RBAC, and additional resources to the shoot cluster via
+  `ManagedResource`.
+
+#### diki-operator
```
timebertt, line 434: The `diki-operator` could also include the admission component instead of deploying another (mostly idling) admission component.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,647 @@
...
+deployed into shoot clusters or their control planes.
+
+This approach was rejected because:
+
+- Diki implements Gardener-aware rules. The diki rulesets understand
+  Gardener-specific architecture (e.g., control plane on seed) and
+  can evaluate rules in this context. Trivy Operator treats every cluster
+  as a generic Kubernetes installation.
+- Trivy Operator does not support the DISA Kubernetes STIG ruleset.
+- No control over the upstream Trivy project. Diki is part of the Gardener
```
timebertt: Generally, such arguments are somewhat true but should never be the driving factor for us (as good open-source citizens). If upstream projects don't implement our feature requests, it's likely not due to a lack of interest or priority but rather due to contributions from our side ;)
On the other hand, statements like this just emphasize the "not invented here" syndrome.

timebertt (commented):
Thanks for the nicely written proposal.
Sorry for the late review, though.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes.

geps/0063-diki-extension/README.md
AleksandarSavchev: Removed manifests with latest commit. https://github.com/gardener/enhancements/pull/64/commits/7e80e32f7507da7dc989069884b2d935139c19a7

geps/0063-diki-extension/README.md
AleksandarSavchev: Nope, it was forgotten 😅

geps/0063-diki-extension/README.md
AleksandarSavchev: The garden type should be prioritised, since it would cover the whole scenario.

The garden runtime cluster scan would cover the runtime cluster workload (data plane) (what would be covered by a seed (soil) diki extension) and the virtual garden scan would cover the virtual garden data plane + control plane

geps/0063-diki-extension/README.md
AleksandarSavchev: We would want when the `ReportOutput` is created and referenced by a `ComplianceScan` to be easily traceable and not be mutated during the execution.

Additionally we can add finalizers to `ReportOutput`s and rule option `ConfigMap`s, when a `{Scheduled}ComplianceScan` references them and remove it when there are no longer referenced.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes, the admission component is part of the `diki-operator`. The responsibilities are added here.

geps/0063-diki-extension/README.md
timebertt: Sounds good. Can you clarify this in the doc?

geps/0063-diki-extension/README.md
timebertt: Ok. In that case, I would update the diagrams because the additional boxes for `diki-admission-controller` suggest that this is an additional component.

timebertt (commented):
Thanks for addressing my feedback. I have no further comments.
Happy to discuss this proposal in this week's steering session :)

geps/0063-diki-extension/README.md
AleksandarSavchev: Added with https://github.com/gardener/enhancements/pull/64/commits/d5b7c5032750eda73d081bbb52b21d7e4f3d0a53

geps/0063-diki-extension/README.md
AleksandarSavchev: Updated the component diagram, but left the sequence diagram, since they are grouped in the same bubble there with https://github.com/gardener/enhancements/pull/64/commits/d5b7c5032750eda73d081bbb52b21d7e4f3d0a53

timebertt (commented):
Thanks for the update. lgtm

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,575 @@
...
+proof of adherence to security baselines such as the DISA Kubernetes STIG.
+
+Currently, Gardener recommends the Diki CLI for running compliance scans. This
+places the full operational burden on each user:
+
+- Manual scheduling: Users must set up their own cron jobs or CI pipelines
+  to trigger scans on a recurring basis.
+- Version management: Users must track Diki CLI releases and update
+  their installations themselves.
+- Configuration management: Each user independently manages Diki
```
rfranzke, line 35: This is still the case even with this proposal, no?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,575 @@
...
+  configuration files, ruleset versions, and rule options.
+- Report collection and storage: Scan results are local JSON files that
```
rfranzke, line 37: Apart from the "storage" part, this is still the same, no? They must read the CRs and do something meaningful with it?

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,575 @@
...
+
+Providing compliance scanning as a first-class Gardener extension removes these
+burdens and makes compliance accessible to all shoot users through familiar
+Kubernetes resource semantics.
+
+### Goals
+
+-  Introduce `gardener-extension-diki` as a Gardener extension
+   that deploys the `diki-operator` into shoot control planes on seeds.
+-  Allow shoot, seed, and garden users to run on-demand compliance scans by
```
rfranzke, line 52: Above point only talks about shoot control planes on seeds, but here you talk about shoot/seed/garden users. This is a bit confusing.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,575 @@
...
+- Scan execution happens on the seed, not in the shoot data plane. The
+  `diki-run` Job runs in the shoot's namespace on the seed. It needs
+  a shoot access secret with the required
+  [RBAC permissions for a Diki scan](https://github.com/gardener/diki/blob/main/example/rbac/managedk8s.yaml)
+  to access the shoot API server. If scan rules need to read the
+  `extensions.gardener.cloud/v1alpha1.Cluster` resource (e.g., for
+  Gardener-specific rulesets), the Job would additionally require access
+  to the seed API server.
+  Additional permissions/credentials would be required, depending on the
+  configured outputs (e.g., credentials to write to a PostgreSQL database).
```
rfranzke, line 115: You could explain here why you don't run the jobs directly in the shoot.

geps/0063-diki-extension/README.md
```diff
@@ -0,0 +1,575 @@
...
+compliance reports. It is immutable and can be referenced by multiple
+`ComplianceScan` resources. Each `ReportOutput` defines exactly one output
+type. The initial implementation supports `ConfigMap`-based output; additional
+outputs will be added in future iterations (e.g., PostgreSQL, OpenSearch).
+
+> [!NOTE]
+> ConfigMap report outputs have an `ownerReference` to the originating
+> `ComplianceScan` and are automatically deleted when the `ComplianceScan` is
+> removed. Reports stored in external systems (e.g., PostgreSQL, OpenSearch)
+> are not cleaned up and remain as-is.
```
rfranzke, line 308: Why? Shouldn't we clean this up when the `ComplianceScan` or `ReportOutput` resources are deleted ?

geps/0063-diki-extension/README.md
ScheererJ: As discussed, this will not be part of the initial scope.

geps/0063-diki-extension/README.md
AleksandarSavchev: Yes and no.

This proposal is in the direction of minimizing the configuration work of users. The operator builds the final diki config file (targeting the cluster, using the provided rule options and ruleset versions)

Additionally with the new diki `config merge` feature from diki [v0.27.0](https://github.com/gardener/diki/releases#release-v0.27.0) we can have installation specific base/ default rule options.

geps/0063-diki-extension/README.md
AleksandarSavchev: Correct.

The collect and store process would be simplified by the `ScheduledComplianceScan` and `ReportOutput` CRDs.

Depending on where they export the compliance findings different approaches can be taken (e.g. ODG handles diki findings by creating a github issue). This has been noted in the non-goals.

geps/0063-diki-extension/README.md
AleksandarSavchev: Improved above point with https://github.com/gardener/enhancements/pull/64/commits/6136e2c89f98e174c40a2d05789c0f40713a2260

geps/0063-diki-extension/README.md
AleksandarSavchev: Added improvements with https://github.com/gardener/enhancements/pull/64/commits/6136e2c89f98e174c40a2d05789c0f40713a2260

geps/0063-diki-extension/README.md
AleksandarSavchev: The generated reports would be needed to show compliance evidence and audit traceability. While the k8s resources `ComplianceScan`s, once completed (reports stored) are advised to be cleaned up.

This would also require the provided storage credentials to have delete permissions.

